PR #4, opened 2026-02-05
chore: GitHub setup

scottschreckengaust:
Setup GitHub repository.

#### Related

- Fixed the `mise ci` for cross validation.
- Added GitHub workflows

#### Changes

* GitHub workflows
* pull request templates
* Security baselines
* CODEOWNERS
* alphabetized JSON
* fixed Markdown linting

#### Acknowledgment

### DIFF
--- a/.github/workflows/security-scanners.yml
+++ b/.github/workflows/security-scanners.yml
@@ -61,7 +61,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
-      - uses: PyCQA/bandit-action@67a458d90fa11fb1463e91e7f4c8f068b5863c7f # v1.0.1
+      - uses: PyCQA/bandit-action@cd2700ff8e8a10b277288e068d0c207c614c46ee # main
         continue-on-error: true
       - uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
         if: always()


DEVELOPMENT_GUIDE.md
krokoko, line 1: This file is not necessary, we will have a proper development guide at the root of the repo

CONTRIBUTING.md
```diff
@@ -2,6 +2,14 @@
 
 Thank you for your interest in contributing to Agent Plugins for AWS.
 
+## Role Guides
```
krokoko, line 5: I would not list the developer guide here

.github/guides/ADMINISTRATIVE_GUIDE.md
```diff
@@ -0,0 +1,38 @@
+# Administrative Guide
```
krokoko: Mise is used to provide a front end for interacting with this repository
These commands should be exposed as mise tasks

DEVELOPMENT_GUIDE.md
scottschreckengaust: Moved `DEVELOPMENT_GUIDE.md` to the top-level.

CONTRIBUTING.md
scottschreckengaust: Still believe a reference to the `DEVELOPMENT_GUIDE.md` is valuable.

.github/guides/ADMINISTRATIVE_GUIDE.md
scottschreckengaust: Removed. Will add tasks as necessary in future PR

PR #16, opened 2026-02-09
chore(doc): update agents.md and security scanners workflow trigger

krokoko:
Update agents.md and remove security scanners workflow from running on each PR since it's already done as part of the build workflow

#### Acknowledgment

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,9 +2,7 @@
 
 ## TL;DR Pitch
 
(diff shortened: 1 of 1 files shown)


.github/workflows/security-scanners.yml
```diff
@@ -3,8 +3,6 @@ on:
   schedule:
     # Daily at 15:12 UTC (random time to avoid GitHub Actions load spikes)
     - cron: '12 15 * * *'
-  push:
-  pull_request:
```
scottschreckengaust, line 7: Can we leave this in until this is part of the quality gates build? We want to prevent security issues from being part of pull_requests (at minimum) I'm fine if it get switched to only pr's to `main`

```suggest
    push:
        - branches: [main]
    pull_request:
        - branches: [main]
```

krokoko: Done in [d95f469](https://github.com/awslabs/agent-plugins/pull/16/commits/d95f4699159b4b1bb095145f3a4d682630fe82fa)

PR #33, opened 2026-02-17
feat(github): add upgrade main workflow

krokoko:
Automated dependency upgrades: add an upgrade-main GitHub workflow that runs on a schedule and on demand, uses mise to upgrade tools with [mise upgrade --bump](https://mise.jdx.dev/cli/upgrade.html#l-bump), captures changes as a patch, and opens a pull request with the updates.

N/A

New workflow .github/workflows/upgrade-main.yml:
- Triggers: workflow_dispatch and daily schedule (00:00 UTC).
- Upgrade job: Checkout main, setup mise (with cache), mise install, then mise upgrade --bump; create a patch from any changes and upload it as an artifact.
- PR job: Runs only when there are changes; downloads the patch, applies it, and uses peter-evans/create-pull-request to open a PR from branch github-actions/upgrade-main with the dependencies label.
- Permissions: contents: read for the upgrade job; contents: write and pull-requests: write for the PR job (using UPGRADE_MAIN_TOKEN PAT token since the default token has read only permissions and we can't update these).
- PR body: Uses the project’s template (summary, Related, Changes, Acknowledgment) and links to the workflow run.

The build workflow with all checks and scanners will run on the created pull request


.github/workflows/upgrade-main.yml
```diff
@@ -0,0 +1,105 @@
...
+  workflow_dispatch: {}
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  upgrade:
+    name: Upgrade
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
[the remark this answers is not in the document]
krokoko: Fixed in latest commit

.github/workflows/upgrade-main.yml
```diff
@@ -0,0 +1,122 @@
...
+          path: ${{ runner.temp }}
+
+      - name: Apply patch
+        run: |
+          [ -s ${{ runner.temp }}/repo.patch ] && git apply ${{ runner.temp }}/repo.patch || echo "Empty patch. Skipping."
+
+      - name: Set git identity
+        run: |-
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
scottschreckengaust, line 93: do we know if this is necessary?

.github/workflows/upgrade-main.yml
```diff
@@ -0,0 +1,122 @@
...
+          body: |-
+            Upgrades project dependencies. See details in [workflow run].
+
+            [Workflow Run]: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+
+            ------
+
+            *Automatically created by the "upgrade-main" workflow*
+          author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
+          committer: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
```
scottschreckengaust, line 120: does this need to be here also? Also, there may need to be an update to the pull request validation for users excluded from contribution statements.

.github/workflows/upgrade-main.yml
krokoko: Required? No. Worth keeping? Yes, if you want the upgrade commits to be clearly from the bot and consistent across forks and upstream. See an example here: https://github.com/awslabs/generative-ai-cdk-constructs/pull/1272
They tell the peter-evans/create-pull-request action which author/committer to set on the commit it creates. If you omit them, the action still creates the commit and the PR; it will use its own defaults (often the user associated with the token).

.github/workflows/upgrade-main.yml
krokoko: See above

scottschreckengaust: Please update the pull request lint workflow along with this PR to prevent validation failures:

https://github.com/awslabs/agent-plugins/blob/main/.github/workflows/pull-request-lint.yml#L55

scottschreckengaust (changes requested):
Add exclusion for upgrade PR for a contributor statement.

.github/workflows/upgrade-main.yml
krokoko: done in [bd57b13](https://github.com/awslabs/agent-plugins/pull/33/commits/bd57b1343bd5747972b8f2b6e3394186ea0e828e)

PR #38, opened 2026-02-18
chore(doc): update guidelines

krokoko:
Update design guidelines to clarify spec compliance with agent skills

#### Acknowledgment

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,25 @@
 
 Agent Plugins for AWS equip AI coding agents with the skills to help you architect, deploy, and operate on AWS. Agent plugins are currently supported by Claude Code and Cursor.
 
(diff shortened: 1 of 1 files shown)


docs/DESIGN_GUIDELINES.md
```diff
@@ -115,77 +115,70 @@ Plugins should have a **clear, cohesive purpose**. Bundle related skills and MCP
...
-tags:
-  - aws
-  - deployment
-  - infrastructure
-  - iac
-  - fargate
-  - lambda
+description: "Deploy to AWS with architecture recommendations and IaC. Triggers on: deploy to AWS, host on AWS, run on AWS. Supports Flask, React, static sites, Fargate, Lambda."
+metadata:
+  tags: aws, deployment, infrastructure, iac, fargate, lambda
```
scottschreckengaust, line 164: has the specification changed from arrays to strings separated by commas?

krokoko: The previous example was wrong. From the Agent Skills specification, the metadata field is a map from string keys to string values. So every value in metadata must be a string. A YAML list like:
```
tags:
  - aws
  - deployment
  - infrastructure
```
is an array, not a string, so it doesn’t fit the spec’s “string values” rule.
Using a single string for tags under metadata is spec-compliant:
```
metadata:
  tags: aws, deployment, infrastructure, iac, fargate, lambda
```
Clients that support metadata.tags can split that string (e.g. on commas or spaces) for search/filtering. So the update from a list to a string under metadata is correct.

dineshSajwan (approved):

PR #85, opened 2026-03-09
feat(migration-to-aws): v2 with 6-phase workflow, AI workload detection, and billing discovery

icarthick:
## Summary

Rewrites the migration-to-aws plugin from a 4-phase to a 6-phase workflow with three parallel discovery paths.

- **6-phase workflow**: discover → clarify → design → estimate → generate → feedback (v1 had discover → clarify → estimate → execute)
- **AI workload detection**: App-code discovery scans for Gemini, OpenAI, Vertex AI, and traditional ML frameworks (TensorFlow, PyTorch), with source-specific model mapping tables (ai-gemini-to-bedrock, ai-openai-to-bedrock)
- **Billing discovery path**: Analyzes GCP billing exports as a fallback when no Terraform or app code is available
- **Adaptive clarify phase**: Category-based questions (global, compute, database, AI, AI-only) that activate based on discover findings
- **Separate design phase**: Dedicated design-infra, design-ai, and design-billing references with confidence levels
- **Generate phase**: Produces Terraform configs, AI provider adapters, Bedrock setup scripts, comparison test harnesses, billing artifacts, and migration documentation
- **Feedback phase**: Anonymized telemetry traces capturing phase timings and complexity metrics (no PII)
- **Pricing cache**: Validated rates with confidence levels; awspricing MCP used for real-time validation

## Changes

- **52 files changed** (+7,697 / -360)
- **37 new files**: phase references (clarify, design, estimate, generate, feedback), design-refs (ai-gemini-to-bedrock, ai-openai-to-bedrock), shared schemas, pricing cache
- **15 modified files**: SKILL.md, plugin.json, README, existing design-refs and discover/clustering references
- All changes scoped to `plugins/migration-to-aws/`

## How to test

```bash
# Install plugin locally
claude --plugin-dir ./plugins/migration-to-aws

# Test against a sample GCP Terraform project
# The skill triggers on: "migrate from GCP", "GCP to AWS", etc.

# Verify build
mise run build
```

## Build status

- Markdown lint: pass
- Manifest validation: pass
- Cross-reference validation: pass
- Formatting (dprint): pass
- Semgrep: informational findings only (detects "OpenAI" keyword in migration reference docs — expected)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -97,6 +97,24 @@
         "step functions"
       ],
       "version": "1.0.0"
+    },
+    {
+      "category": "fullstack",
+      "description": "Build full-stack apps with AWS Amplify Gen 2 using guided workflows for authentication, data models, storage, GraphQL APIs, and Lambda functions.",
+      "keywords": [
+        "aws",
+        "amplify",
+        "fullstack",
+        "authentication",
+        "data",
+        "storage",
+        "graphql",
+        "functions"
+      ],
+      "name": "aws-amplify",
+      "source": "./plugins/aws-amplify",
+      "tags": ["aws", "amplify", "fullstack"],
+      "version": "1.0.0"
     }
   ]
 }
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -31,6 +31,7 @@ tools/                              @awslabs/agent-plugins-admins
 ## Plugins (alphabetically listed)
 
 plugins/amazon-location-service     @awslabs/agent-plugins-admins @awslabs/agent-plugins-maintainers  @awslabs/agent-plugins-amazon-location-service
+plugins/aws-amplify                 @awslabs/agent-plugins-admins @awslabs/agent-plugins-maintainers  @awslabs/agent-plugins-amplify
 plugins/aws-serverless              @awslabs/agent-plugins-admins @awslabs/agent-plugins-maintainers  @awslabs/agent-plugins-aws-serverless
 plugins/deploy-on-aws               @awslabs/agent-plugins-admins @awslabs/agent-plugins-maintainers  @awslabs/agent-plugins-deploy-on-aws
 plugins/migration-to-aws            @awslabs/agent-plugins-admins @awslabs/agent-plugins-maintainers  @awslabs/agent-plugins-migrate-to-aws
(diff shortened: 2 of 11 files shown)


plugins/migration-to-aws/skills/gcp-to-aws/references/phases/discover/discover.md
```diff
@@ -1,109 +1,178 @@
...
+- If not found → Skip. Log: "No billing files found — skipping billing discovery."
+- If found AND **no** Terraform files from 1a → Load `references/phases/discover/discover-billing.md` (billing is the primary source — needs full processing for the billing-only design path).
+- If found AND Terraform files **were** found in 1a → Use lightweight extraction below. Do **not** load `discover-billing.md`.
+
+**Lightweight billing extraction (when IaC is the primary source):**
+
+When Terraform is present, billing data is supplementary — only service-level costs and AI signal detection are needed. Extract via a script to avoid reading the raw file into context.
+
+1. Use Bash to read only the **first line** of the billing file to identify column headers.
+2. Write a script to `$MIGRATION_DIR/_extract_billing.py` (or `.js` / shell — use whatever runtime is available) that:
```
krokoko, line 81: This path triggers when both Terraform files and billing files are present. I have some concerns here since the python script is LLM-generated. The skill instructions tell the LLM to write a script that parses the billing file. The exact code varies per invocation, it depends on:
  - The column headers the LLM reads from line 1
  - Which runtime the LLM decides to use (Python, Node, shell)
  - The LLM's interpretation of the instructions
This means there's no fixed, reviewable, testable artifact. Two runs of the same skill on the same billing file could produce different scripts. If the script has a bug (e.g., wrong column index, float parsing error), there's no way to reproduce it because the script is deleted in step 5. 

The LLM reads the first line of the CSV (column headers) and uses those headers to write the script. If the billing file has unusual or adversarial column names, the LLM might generate a script that behaves unexpectedly.

Step 3 says: "try python3 first. If not found, try python. If neither available, delete the script and fall back to loading discover-billing.md." This means the skill might generate Python code, fail to execute it, then fall back to having the LLM read the file directly (the thing it was trying to avoid). The fallback silently changes the behavior and context cost.

Ans as hinted above: Step 5 says "Delete the script file after successful execution." This means:
  - If the output is wrong, you can't inspect what script produced it
  - If there's a security incident, the executed code is gone
  - You can't diff the script between runs to understand behavioral changes

I would have either:
- have a versioned fixed extraction script, for instance: Add a tools/extract-billing.py (or similar) to the repo. It takes a billing CSV path as input, outputs the billing-profile.json schema to stdout. The skill instructions say "run tools/extract-billing.py <path>" instead of "write a script." This is reviewable, testable, and deterministic.
- or a simpler heuristic without code generation

icarthick: Yes I agree a script here would work well as opposed to instructions. For complex billing data, LLM automatically resorts to python script generation, regardless of whether i have instructions to create python script or not. Billing data invariably requires some kind of math, which i feel simple instructions might not be able to capture

plugins/migration-to-aws/.claude-plugin/plugin.json
scottschreckengaust: Please update this based on semantic versioning.

PR #121, opened 2026-04-07
fix: correct nosemgrep rule IDs for sagemaker-ai and tools

scottschreckengaust:
## Summary

- Fix `nosemgrep` inline comments that used truncated rule IDs, causing suppressions to be silently ignored by semgrep
- All 7 findings are false positives with explanatory comments preserved:
  - `format_detector.py` (3 findings): `is_valid` is a `@dataclass` `bool` field, not a method — `-- dataclass field, not a method`
  - `validate-cross-refs.cjs` (4 findings): `marketplacePath` parameter is only ever called with hardcoded string constants — `-- callers pass hardcoded constants, not user input`

## Test plan

- [ ] Run `mise run security:semgrep` — should show 0 findings (down from 7 in these files)
- [ ] Run `mise run build` — passes cleanly

---

### DIFF
--- a/plugins/sagemaker-ai/skills/dataset-evaluation/scripts/format_detector.py
+++ b/plugins/sagemaker-ai/skills/dataset-evaluation/scripts/format_detector.py
@@ -653,7 +653,7 @@ if __name__ == "__main__":
         if args.json:
             output = {
                 "format_type": result.format_type.value,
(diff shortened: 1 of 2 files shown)


krokoko (approved):

mayakost (approved):
LGTM

PR #132, opened 2026-04-15
fix(deploy-on-aws): check file extension before defusedxml dependency

### LINKED ISSUE #129: deploy-on-aws: validate-drawio hook fires dependency warning on every file write, not just .drawio files
## Problem

The `deploy-on-aws` plugin registers a `PostToolUse` hook on `Edit|Write` that runs `validate-drawio.sh`. This script checks for the `defusedxml` Python dependency **before** checking whether the file being written is actually a `.drawio` file.

As a result, every single file write (HTML, markdown, JSON, memory files, etc.) produces:

```
Missing required dependency: defusedxml. Install it with: pip3 install defusedxml>=0.7.1
```

...even though the hook has nothing to do with the file being written.

## Location

`scripts/validate-drawio.sh`, lines 13-16:

```bash
python3 -c "import defusedxml" 2>/dev/null || {
  echo '{"systemMessage": "Missing required dependency: defusedxml. Install it with: pip3 install defusedxml>=0.7.1"}'
  exit 0
}
```

The file extension check comes later in the script. The dependency check should either:
1. Come **after** the file type check (so non-drawio files exit early and silently), or
2. The hook matcher in `hooks.json` should be scoped to `.drawio` files rather than matching all `Edit|Write` events

## Impact

Anyone with the plugin enabled who hasn't installed `defusedxml` sees this warning on every file write across every project and file type. Produces noise that undermines trust in the hook system.

## Version

`deploy-on-aws` v1.2.0

Sagargupta16:
## Summary

- Move the `defusedxml` dependency check after the file extension filter in `validate-drawio.sh`
- Non-drawio files now exit early and silently instead of showing a dependency warning on every Edit/Write

## Problem

The `PostToolUse` hook on `Edit|Write` runs `validate-drawio.sh` for every file write. The script checked for the `defusedxml` dependency before checking whether the file was a `.drawio` file, causing every non-drawio write (markdown, JSON, etc.) to emit:

```
Missing required dependency: defusedxml. Install it with: pip3 install defusedxml>=0.7.1
```

## Fix

Reordered the script to:
1. Read stdin and extract file path
2. Check file extension (exit early for non-drawio files)
3. Check `defusedxml` dependency (only reached for actual drawio files)

#### Related

Fixes #129

#### Changes

Reordered `validate-drawio.sh` to check file extension before checking the `defusedxml` dependency, so non-drawio file writes exit early and silently.

#### Acknowledgment


krokoko (approved):

scottschreckengaust (approved):
LGTM

PR #141, opened 2026-04-21
feat(databases-on-aws): add query-plan-explainability workflow to dsql skill

anwesham-lab:
## Summary

- Adds **Workflow 8: Query Plan Explainability** to the `dsql` skill under the `databases-on-aws` plugin
- Phased workflow: required-reads (Phase 0) → capture plan → gather evidence → conditional experiments → produce structured diagnostic report
- Four new reference files encapsulate the details (`plan-interpretation.md`, `catalog-queries.md`, `guc-experiments.md`, `report-format.md`) so `SKILL.md` stays under the 300-line guidance
- **Fallback path when the MCP server isn't available** — the workflow routes to `scripts/psql-connect.sh --command` which the plugin already ships
- Adds a functional eval harness at `tools/evals/databases-on-aws/` — 5 scenarios / 42 assertions — driving `claude -p` against a live cluster via `--mcp-config`, with a uvx warmup that aborts on repeated failure to prevent false-negative eval runs
- Also in this PR:
  - **`trigger_evals.json`**: 3 new should-trigger explainability queries + 3 should-not-trigger PostgreSQL/RDS/Redshift EXPLAIN variants to confirm the skill stays DSQL-specific (not covered by the existing trigger suite)
  - **`mise.toml`**: Bandit scan scoped to skip `./.tmp` and `./node_modules` so per-developer scratch code doesn't break the build

## What's in the workflow

- EXPLAIN ANALYZE VERBOSE capture (`readonly_query` for SELECT, `transact` with trailing `ROLLBACK` for DML)
- Catalog interrogation (`pg_class`, `pg_stats`, `pg_indexes`, `COUNT(*)`, `COUNT(DISTINCT)`) with estimation-error classification (2x-5x minor, 5x-50x significant, 50x+ severe per `plan-interpretation.md`)
- Correlated-predicate + data-skew detection
- Composite-index recommendations using `CREATE INDEX ASYNC`
- GUC experiments gated on execution time (≤30s runs experiments; >30s skips and includes manual testing SQL in the report)
- Anomalous-EXPLAIN-value handling (confirm query results are correct despite inflated counters, produce a Support Request Template)
- Structured Markdown report with a fixed template (Query Information, SQL Statement, Plan Overview, numbered Findings ordered by Node Duration, Summary table)

## Why integrated into `dsql` rather than a standalone skill

This started as a standalone `dsql-query-troubleshooting` skill. Trigger evals showed the standalone description barely activated the skill (1/10 queries triggered it). Integrating into `dsql` — which already triggers on any Aurora DSQL mention — gave the workflow full context (DSQL conventions like `CREATE INDEX ASYNC`, `pg_indexes` discovery, transaction limits) and better trigger coverage. The v3 functional eval pass rate jumped from 56% (standalone) to 76% (integrated) on the same assertions.

## What's intentionally NOT in this PR

- **No fixture seeds** — each eval's `README.md` documents the schema shapes it expects, but seeding the cluster is left to the operator. A reference seed script lives under `.tmp/` (gitignored) and can be shared separately if useful.
- **No MCP config for the eval cluster** — operators supply their own via `--mcp-config`. The plugin's shipped `.mcp.json` keeps `aurora-dsql` disabled by default.

## Test plan

- [x] `mise run build` passes (lint + fmt + validate + security scans)
- [x] Functional evals run end-to-end against a live cluster with seed fixtures
- [x] MCP warmup handles cold uvx/boto3 startup; aborts on repeated failure so evals never run against an uninitialized server
- [x] Tool-required assertions require actual MCP tool calls (no prompt-keyword false-positives)
- [ ] Follow-up: further tighten report-format adherence (agent loads `report-format.md` but doesn't always mirror the H4 structure verbatim — iterating separately)

---


plugins/databases-on-aws/skills/dsql/references/query-plan/guc-experiments.md
```diff
@@ -0,0 +1,159 @@
...
+
+---
+
+## GUC Experiment Procedure
+
+GUC (Grand Unified Configuration) experiments temporarily disable specific planner strategies to test whether viable alternatives exist.
+
+### Experiments to Run
+
+Execute two variants against the **original SQL statement** (unmodified). `transact` auto-wraps its statement list in a single `BEGIN/COMMIT`, so do **not** include `BEGIN` or `COMMIT` in the list. If the original query is DML, append `ROLLBACK` as the final statement per SKILL.md Phase 1 so nothing persists.
```
Morlej: For safety, we probably want to reject any query is DML. Only UPDATE and DELETE can have query plan issues, and both of those can be rewritten as plan SELECTs. We could also do that conversion and just do the EXPLAIN ANALYZE on the resulting SELECT. INSERTs should be rejected, as should any kind of procedural language such as pl/pgsql

plugins/databases-on-aws/skills/dsql/references/query-plan/guc-experiments.md
```diff
@@ -0,0 +1,159 @@
...
+```
+
+**Experiment 2 — Merge join only** (needs SETs, so use `transact`):
+
+```python
+transact([
+    "SET LOCAL enable_hashjoin = off",
+    "SET LOCAL enable_nestloop = off",
+    "SET LOCAL enable_mergejoin = on",
+    "EXPLAIN ANALYZE VERBOSE {original_sql}",
```
Morlej, line 39: Ouch, further to my point above, we need to make sure that customers can't foot gun by way of SQL injecting themselves. Does the `readonly_query` role not allow for SETting of GUCs?

plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md
```diff
@@ -0,0 +1,188 @@
...
+### DSQL-Specific Nodes
+
+| Node Type               | Description                                                                 |
+| ----------------------- | --------------------------------------------------------------------------- |
+| Full Scan (btree-table) | Logical full table scan decomposed into Storage Scan + B-Tree Scan children |
+| Storage Scan            | Physical read of row data from storage layer                                |
+| B-Tree Scan             | Physical read of index entries from B-Tree storage                          |
+| Storage Lookup          | Point lookup of a row by internal row pointer (follows index scan)          |
+| B-Tree Lookup           | Point lookup of an index entry by key                                       |
+
```
Morlej, line 31: Replace the above table with this:

DSQL table data is all stored in B-Tree structures. Secondary indexes are also B-Tree, and contain the primary table keys for the secondary index values that make up the tree. 

Full Scan (btree-table) — Scans the entire table without using an index. This is DSQL's equivalent of a sequential scan (Seq Scan) in standard PostgreSQL. It reads all rows in the table's B-Tree key range and is chosen when no index is selective enough to be worthwhile, or when no applicable index exists for the query's filter conditions. Like other scan nodes, it nests a Storage Scan / B-Tree Scan pair underneath it for projection pushdown, filter pushdown, and limit pushdown to storage.

Storage Scan — Reads rows from storage by scanning a key range. This is the logical scan operator that applies server-side projection (column filtering), predicate filtering, and limit pushdown before returning results to the query processor. It appears in sequential scans, index scans, index-only scans, and bitmap index scans.

B-Tree Scan — The physical read operation underneath a Storage Scan. It traverses the B-Tree key range in storage to produce candidate rows, which the parent Storage Scan then filters. The "Index Cond" shown beneath it indicates which key conditions bound the scan range.

Storage Lookup — Fetches specific rows from the primary table by key, typically after a secondary index scan identifies which rows are needed. It applies projection and filter pushdown on the table data. Appears in secondary index scans (where the index scan finds keys, then this node fetches the actual table rows) and bitmap heap scans (where a bitmap of keys is resolved to table rows).

B-Tree Lookup — The physical point-read operation underneath a Storage Lookup. It retrieves individual rows from the B-Tree by their primary key, as opposed to scanning a range. This is the table-side counterpart to B-Tree Scan.

Buffer — A batching node used during INSERT ... ON CONFLICT operations. It collects tuples from its child plan into batches, then probes all relevant unique/arbiter indexes in bulk to pre-populate a conflict cache. This avoids per-row round-trips to storage for conflict detection. Shows "Conflict Resolution" (NOTHING or UPDATE) and "Conflict Arbiter Indexes" in its EXPLAIN output.

Filters — (Sub-property, not a node) Predicates that have been pushed down to the storage layer for server-side evaluation, reducing the amount of data transferred back to the query processor. Shown with "Rows Removed by Filter" in EXPLAIN ANALYZE.

Limit — (Sub-property, not a node) When a SQL LIMIT clause is present, the limit value is pushed down to storage so it can stop scanning early rather than returning all matching rows.

Index Cond — (Sub-property, not a node) The key conditions used to bound the B-Tree scan or lookup range. These determine the start and end points of the key interval being read from storage.

Statement DPU Estimate (shown in EXPLAIN ANALYZE VERBOSE) — A summary section appended at the end of the plan showing the estimated Distributed Processing Unit (DPU) cost breakdown for the statement. It includes Compute, Read, Write, and optionally Multi-Region Write costs, plus a Total. This helps you understand the resource consumption and billing impact of a query.

Projections (sub-property on Storage Scan, Storage Lookup, and Full Scan nodes) — Lists the specific columns being pushed down to storage for retrieval. Only the listed columns are read from storage, reducing I/O. When all columns are optimized away (e.g., COUNT(*) queries), it shows <All columns optimized out>.

Query Identifier — Shown in DSQL's EXPLAIN output as a base32-encoded string (rather than the standard PostgreSQL bigint). This identifier correlates the query plan with entries in pg_stat_statements and other monitoring views.

plugins/databases-on-aws/skills/dsql/references/query-plan/guc-experiments.md
anwesham-lab: Good call — pushed `858bcd0`. SKILL.md 

Phase 1 now rejects `INSERT` and `pl`/`pgsql`/`DO`-blocks/functions outright, and rewrites `UPDATE`/`DELETE` to their equivalent `SELECT` form (same join chain + `WHERE`) before plan capture. 

The optimizer picks the same plan shape for `UPDATE … WHERE …` as for `SELECT … WHERE …` over the same tables, so analysis on the rewritten SELECT is valid. 

Plan capture is now exclusively `readonly_query` — never `transact --allow-writes`, which bypasses all MCP safety checks. Updated `guc-experiments.md` to reference that the statement reaching Phase 3 is always a `SELECT`.

plugins/databases-on-aws/skills/dsql/references/query-plan/guc-experiments.md
anwesham-lab: Checked the MCP server source. `readonly_query` blocks multi-statement (semicolon guard) and blocks `SET GLOBAL`/`SET SESSION`/`SET autocommit`, but `SET LOCAL` isn't in the mutable-keyword set. The multi-statement guard means we can't do `SET LOCAL …; EXPLAIN …` through `readonly_query` — we still need `transact` for that path, which unfortunately does bypass validation in `--allow-writes` mode.

Rather than route around that, I added explicit caller-side rules to the Experiment 2 block: `{original_sql}` must be verified as a SELECT first (Phase 1 guarantees that since DML is rejected/rewritten), the statement must be passed as a single list element (no string concat that could smuggle a second statement), and failure must halt without recovery-chaining. That's the best we can do at the skill layer short of a server-side change.

Happy to open a separate issue against `awslabs/mcp` to alter for the `readonly_set_local` carve-out if you think that's the right direction. Refs: 
* https://github.com/awslabs/mcp/blob/main/src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
* https://github.com/awslabs/mcp/blob/main/src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/server.py

plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md
anwesham-lab: Applied your table verbatim in `858bcd0`, plus added the B-Tree storage context ("DSQL stores all table data in B-Tree structures. Secondary indexes are also B-Tree, and contain the primary table keys for the secondary index values that make up the tree.") as a section intro. Thanks for the correction — when I let the AI run with the guidance, the original descriptions conflated the logical "Full Scan" node with the physical children.

plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md
```diff
@@ -0,0 +1,188 @@
...
+| Node Type   | Description                                            |
+| ----------- | ------------------------------------------------------ |
+| Nested Loop | Iterates inner side once per outer row                 |
+| Hash Join   | Builds hash table from one side, probes with the other |
+| Merge Join  | Merges two pre-sorted inputs                           |
+| Index Scan  | Scans an index and fetches matching rows               |
+| Seq Scan    | Sequential full table scan                             |
+| Sort        | Sorts rows for Merge Join or ORDER BY                  |
+| Aggregate   | Computes GROUP BY / aggregate functions                |
+
```
Morlej, line 44: Add:
| Index Only Scan | Retrieves all data from index access (no table access) |

plugins/databases-on-aws/skills/dsql/SKILL.md
```diff
@@ -250,16 +252,42 @@ MUST load [ddl-migrations/overview.md](references/ddl-migrations/overview.md) be
...
+
+**Phase 0 — Load reference material.** Read all four before starting — each has content later phases need verbatim (node-type math, exact catalog SQL, the `>30s` skip protocol, required report elements):
+
+1. [query-plan/plan-interpretation.md](references/query-plan/plan-interpretation.md) — node types, duration math, anomalous values
+2. [query-plan/catalog-queries.md](references/query-plan/catalog-queries.md) — pg_class / pg_stats / pg_indexes SQL
+3. [query-plan/guc-experiments.md](references/query-plan/guc-experiments.md) — GUC procedures and `>30s` skip protocol
+4. [query-plan/report-format.md](references/query-plan/report-format.md) — required report structure
+
+**Phase 1 — Capture the plan.** Execute the plan capture immediately — the user provided the SQL precisely so you can run it, and the read-only `EXPLAIN ANALYZE` path is safe by construction. For SELECT, run `EXPLAIN ANALYZE VERBOSE …` via `readonly_query`. For DML, use `transact(["EXPLAIN ANALYZE VERBOSE …", "ROLLBACK"])` — the trailing ROLLBACK prevents persistence. Extract Query ID, Planning Time, Execution Time, DPU Estimate.
+
```
Morlej, line 267: See note above regarding DML safety

tools/evals/databases-on-aws/README.md
```diff
@@ -64,3 +67,75 @@ PYTHONPATH="<skill-creator-path>:$PYTHONPATH" python -m scripts.run_loop \
...
+| ---- | ---------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| 1    | `data.Associate`, `data.AssociateIndicative`, `data.WorkAgreement`           | Bi-temporal PK `(AssociateID, _transactionstartdatetime)`. Seed ~50 rows under the target `ClientID = 'a19fa90e-6b95-416b-9d69-60c4809c31c5'` plus ~950 decoys across other clients. All rows share `_transactionstartdatetime = '2000-01-01 00:00:00.000'`. Create ONLY a single-column index on `_transactionstartdatetime` so the composite-index recommendation has a real gap to surface. |
```
Morlej: These are customer specific and we should remove

Morlej (commented):
Looks good once my comments are addressed/incorporated. Once landed we can test with many more use cases.

plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md
anwesham-lab: Added in `65566cd`.

plugins/databases-on-aws/skills/dsql/SKILL.md
anwesham-lab: Covered in `858bcd0` — Phase 1 now rejects INSERT / pl-pgsql / DO-blocks / functions outright and rewrites UPDATE/DELETE to the equivalent SELECT (same join chain + WHERE) before running EXPLAIN ANALYZE VERBOSE via `readonly_query`. Eval 4 was also updated to exercise the UPDATE-rewrite path instead of the old INSERT-ROLLBACK wrapping.

tools/evals/databases-on-aws/README.md
anwesham-lab: Removed in `65566cd`. Eval 1 prompt and the README fixture table now use generic `public.user_account` / `user_profile` / `work_assignment` schemas with a neutral `tenant_id` UUID (`11111111-2222-3333-4444-555555555555`). The matching predicate-inference example in `guc-experiments.md` was also genericized ("all orders for a user belong to the same tenant as the user").

plugins/databases-on-aws/skills/dsql/SKILL.md
```diff
@@ -250,16 +252,42 @@ MUST load [ddl-migrations/overview.md](references/ddl-migrations/overview.md) be
...
+PGPASSWORD="$TOKEN" psql "host=$HOST port=5432 user=admin dbname=postgres sslmode=require" <<<"EXPLAIN ANALYZE VERBOSE <sql>;"
+```
+
+**Safety.** Plan capture uses `readonly_query` exclusively — it rejects INSERT/UPDATE/DELETE/DDL at the MCP layer. Rewrite DML to SELECT (Phase 1) rather than asking `transact --allow-writes` to run it; write-mode `transact` bypasses all MCP safety checks. **MUST NOT** run arbitrary DDL/DML or pl/pgsql.
+
 ---
 
 ## Error Scenarios
 
-- **MCP server unavailable:** Fall back to CLI scripts ([scripts/](../../scripts/)) or direct psql. Note the limitation to the user.
```
amaksimo, line 257: Why did we remove this?

tools/evals/databases-on-aws/query_explainability_evals.json
```diff
@@ -0,0 +1,86 @@
...
+        "Recommends a composite index covering both customer_id and status",
+        "Uses CREATE INDEX ASYNC syntax in the recommendation",
+        "Produces a structured diagnostic report",
+        "Does NOT execute any DDL or DML statements"
+      ]
+    },
+    {
+      "id": 3,
+      "prompt": "I have a DSQL query joining 4 tables that takes 45 seconds. Can you help me figure out why it's so slow?\n\nSELECT e.name, d.dept_name, p.project_name, t.hours FROM employees e JOIN departments d ON e.dept_id = d.id JOIN project_assignments pa ON e.id = pa.employee_id JOIN projects p ON pa.project_id = p.id JOIN timesheets t ON e.id = t.employee_id WHERE d.location = 'NYC' AND t.week_start = '2024-01-15';\n\nI've tried adding indexes but nothing seems to help.",
+      "expected_output": "Executes EXPLAIN ANALYZE VERBOSE, identifies expensive nodes, but skips GUC experiments due to >30s execution time, recommends manual testing",
```
amaksimo, line 44: nit: 45? not 30?

tools/evals/databases-on-aws/trigger_evals.json
```diff
@@ -39,10 +39,34 @@
     "query": "help me write a query to get all orders for a specific tenant in my distributed sql database on aws, making sure we handle the sql injection risk properly",
     "should_trigger": true
   },
+  {
```
amaksimo, line 42: These types of tests are super cool - I was not aware you could do something like that, TIL!

tools/evals/databases-on-aws/README.md
```diff
@@ -64,3 +67,75 @@ PYTHONPATH="<skill-creator-path>:$PYTHONPATH" python -m scripts.run_loop \
...
+`aurora-dsql` disabled by default. Supply your own MCP config via `--mcp-config`, pointing to
+a JSON file with cluster credentials (e.g., `.claude/.mcp.json`, gitignored).
+
+The cluster also needs the schemas and data the eval prompts reference — see
+[Cluster fixtures the evals expect](#cluster-fixtures-the-evals-expect) below. Seeding is
+left to the operator so you can use whatever method fits your environment (psycopg + IAM
+token, `psql`, CDK migrations, etc.).
+
+The runner fires one throwaway `claude -p "hi"` as a uvx warmup before the real evals
+— otherwise the first eval often reports the MCP as "not connected" because `uvx` is
```
amaksimo, line 88: nit: Should we rephrase this to runner starts with checking MCP connectivity by starting a claude session via uvx?

amaksimo (dismissed):

tools/evals/databases-on-aws/README.md
anwesham-lab: it's not checking connectivity, it's forcing connectivity by basically advance calling the UVX install and setting the cache, hence my phrasing accordingly but I see the confusion.

tools/evals/databases-on-aws/query_explainability_evals.json
anwesham-lab: oh I did that when I bloated cluster size, tbh it can be anything, just depends on what you set cluster to. I had done this to get to the cusp of OOM'ing, but I can consolidate docs.

plugins/databases-on-aws/skills/dsql/SKILL.md
anwesham-lab: Good question tbh, I meant to comment explaining this proactively. 

It was bloating length, tested the agent's behavior and it does this regardless of the messaging being there in the SKILL explicitly, and it was falling back to the psql connect script in cases where it should use direct psql when this message was there. I added explicit fallback directives on the workflows anyways so this was redundant.

Morlej (dismissed):

amaksimo (dismissed):

Morlej (approved):

pkale (approved):

theagenticguy (approved):
LGTM

PR #155, opened 2026-05-04
fix(databases-on-aws): correct DSQL type guidance, add cluster-lifecycle troubleshooting

anwesham-lab:
## Summary

- Correct DSQL type guidance in the `databases-on-aws` skill: `JSON` is a supported column type (1 MiB, auto-compressed); `JSONB`, arrays, and `INET` remain [runtime-only](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-supported-data-types.html#working-with-postgresql-compatibility-query-runtime). Quick-reference type lists are replaced with pointers to the canonical [AWS supported data types doc](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-supported-data-types.html) plus an `awsknowledge` verify row, so the skill does not drift as DSQL's type surface evolves.
- Add two entries to `troubleshooting.md` under a new Cluster Lifecycle section: the `FATAL: unable to accept connection, waking up cluster` error emitted when connecting to an `INACTIVE` cluster, and the `FailedPrecondition` returned when backing up an `IDLE`/`INACTIVE` cluster. Links to the [cluster lifecycle docs](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/cluster-lifecycle.html).
- Remove stale `JSON.stringify(...)` wrapping in `data-operations.md` examples now that `metadata` is a `JSON` column.
- Extend the Tier 2 functional eval suite with three new evals covering the updated behaviors (JSON column storage, array-column rejection, INACTIVE-cluster wake flow), plus the grader clauses they need. Add `--eval-ids` to the runner so subsets can be run without executing the full suite.

## Test plan

- [x] Verified against a live DSQL cluster in `us-west-2`:
  - `CREATE TABLE ... (payload JSON)` → accepted; `information_schema` reports `data_type = json`
  - `CREATE TABLE ... (payload JSONB)` → rejected with `ERROR: datatype jsonb not supported`
  - `CREATE TABLE ... (tags TEXT[])` → rejected with `ERROR: datatype text[] not supported`
  - `payload::jsonb->>'key'`, `payload::jsonb @> '{...}'::jsonb`, `string_to_array(...)` all succeed
- [x] Verified lifecycle behavior against a live cluster: `INACTIVE` cluster returns the documented FATAL error on first connection and reaches `ACTIVE` after polling; `IDLE` cluster wakes transparently
[...]

PR #162, opened 2026-05-08
feat(dsql): enhance query plan explainability with type coercion detection, rewrites, and workflow extraction

Morlej:
## Summary

- **Extract Workflow 8** from SKILL.md into `references/query-plan/workflow.md` (SKILL.md: 246 → 249 LOC)
- **Add type coercion index bypass detection** — pg_amop-based detection in `plan-interpretation.md`, indexed column type queries in `catalog-queries.md`
- **Add query rewrite references** — 10 generic patterns split into individual files under `query-rewrites/`, plus 2 DSQL-specific rewrites (reltuples estimate, split large joins)
- **Add structured trigger criteria, context disambiguation, and routing** to the workflow reference
- **Wire rewrites into workflow** — loaded at Phase 0, applied at Phase 2

## Validation

- `validate-size.py`: 249 lines (good, under 300 limit)
- `validate-references.py`: 0 broken links, 0 new orphans

## Eval Results

Manual qualitative comparison (n=1, Claude Opus 4.6). Full results in `tools/evals/databases-on-aws/dsql/query_plan_rewrite_eval_results.md`:

| Eval | Scenario | With Skill | Baseline | Key Delta |
|------|----------|-----------|----------|-----------|
| 200 | IN-subquery Full Scan | PASS | PARTIAL | Skill recommends specific rewrite patterns from reference |
| 201 | Type coercion index bypass | PASS | PASS | Both identify it; skill adds DSQL-specific pg_amop detail |
| 202 | 12-table join ordering | PASS | PARTIAL | Skill offers full diagnostic workflow with GUC experiments |
| 203 | COUNT(*) timeout | PASS | FAIL | Skill recommends pg_class reltuples with staleness warning |
| 204 | Multiple OR to IN | PASS | PARTIAL | Skill identifies pattern from reference |
| 205 | GROUP BY after JOIN | PASS | PARTIAL | Skill recommends subquery aggregation |
| 206–210 | LEFT JOIN, computation push, NOT IN+NULL, UNION ALL, negative | Added in review round | — | Coverage for remaining patterns + negative case |

## Follow-ups

- **MCP mirror PR:** `awslabs/mcp` `src/aurora-dsql-mcp-server/skills/dsql-skill/` needs to be synced with these changes (workflow.md, query-rewrites/ split, updated catalog-queries.md, plan-interpretation.md). Will open companion PR after this merges.
- **Python SQL converter:** Per review feedback, deterministic rewrites should migrate to a Python script in a future PR (reference files then document the converter's rules).


plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites-generic.md
```diff
@@ -0,0 +1,594 @@
...
+**Do not apply:** The IN list is a small static set of constants.
+
+```sql
+-- Original
+SELECT *
+FROM customers
+WHERE customer_id IN (
+  SELECT customer_id
+  FROM orders
+  WHERE order_date >= DATEADD(day, -30, GETDATE())
```
amaksimo: This looks like sql-server syntax?

I thought dsql may reject it, so I tested it locally:
```
ERROR:  function dateadd(unknown, integer, timestamp with time zone) 
does not exist                                                       LINE 1: SELECT DATEADD('day', -30, NOW());
               ^
HINT:  No function matches the given name and argument types. You mig
ht need to add explicit type casts.     
```

The same comment for line 385

plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites-dsql-specific.md
```diff
@@ -0,0 +1,91 @@
...
+
+---
+
+## Replace COUNT(*) with reltuples Estimate
+
+When a query performs `COUNT(*)` on a large table, rewrite to use the `reltuples` value from `pg_class` for an approximate row count. This is a common workaround for cases where `COUNT(*)` is too slow or times out on large tables.
+
+**When to apply:** An approximate count is acceptable and the table is large enough that `COUNT(*)` is prohibitively expensive.
+
+**Do not apply:** The application requires an exact count.
```
amaksimo: This type of language appears quite often, try to use positive language like: "Apply only when..." rather than "Do not apply when..."

plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites-generic.md
```diff
@@ -0,0 +1,594 @@
+# Query Rewrites Reference
```
amaksimo: This is a very long file, it will consume a lot of context once the agent loads it in.

I think we should split it up into sub-files to make it easier for the agent to reference.

Aside, let's use RFC language here and in the dsql-specific file above.

amaksimo (commented):
I have a few general commets:
1. We should use positive language throughout (llm can confuse DO with DO NOT when we trim context)
2. We should try to use RFC language more frequently throughout
3. We should break up the references in the query-plan folder as some of the files are very long

plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md
```diff
@@ -0,0 +1,136 @@
...
+## Phase 0 — Load Reference Material
+
+Read all files before starting — each has content later phases need verbatim (node-type math, exact catalog SQL, the `>30s` skip protocol, required report elements):
+
+1. [plan-interpretation.md](plan-interpretation.md) — node types, duration math, anomalous values
+2. [catalog-queries.md](catalog-queries.md) — pg_class / pg_stats / pg_indexes SQL
+3. [guc-experiments.md](guc-experiments.md) — GUC procedures and `>30s` skip protocol
+4. [report-format.md](report-format.md) — required report structure
+5. [query-rewrites-generic.md](query-rewrites-generic.md) — generic SQL rewrite patterns
+6. [query-rewrites-dsql-specific.md](query-rewrites-dsql-specific.md) — DSQL-specific rewrites
```
amaksimo: Why do we need both dsql-specific rewrites and generic patterns?

Given this skill is for dsql, I am not sure this makes sense?

Aside: we are loading all these files, some are very big. We should probably try to split them up and only load sections (or better yet - individual split files) when they are needed to keep context usage down.

plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md
```diff
@@ -0,0 +1,136 @@
...
+
+End with the "Next Steps" block from that reference so the user can ask for a reassessment after applying a recommendation.
+
+When the user says "reassess" (or equivalent), re-run Phase 1–2 and **append an "Addendum: After-Change Performance"** to the original report (before/after table, match against expected impact) rather than producing a new report.
+
+If a query rewrite was identified in Phase 2, include it as a recommendation with the original and rewritten SQL side by side.
+
+---
+
+## psql Fallback
```
amaksimo: Do we want this? Why not enforce using the mcp? Aside: does this belong here or should we bring that up to the table where we mention it initially?

plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md
```diff
@@ -183,6 +183,63 @@ Detect physically impossible row counts in DSQL plan nodes:
 
 These anomalous values do not affect query correctness — only diagnostic output accuracy.
 
+## Type Coercion and Index Bypass
```
amaksimo, line 221: RFC language?

plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites-generic.md
Morlej: Good catch. This is a hangover from the source document.

plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites-dsql-specific.md
Morlej: Fixed throughout in next rev

plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites-generic.md
Morlej: I split the file and updated for RFC language

plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md
Morlej: We do need both, since the generic ones are generic-but-really-help-dsql, rather than exercising DSQL-specific features/traits

plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md
Morlej: Removed

plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md
Morlej: Done

plugins/databases-on-aws/skills/dsql/SKILL.md
```diff
@@ -164,16 +164,18 @@ defaults that may change — when a user's decision depends on an exact limit, v
...
-| Supported column data types    | See docs      | `aurora dsql supported data types` |
 
-**When to verify:** Before recommending batch sizes, connection pool settings, or schema designs where hitting a limit would cause failures; any time the exact number can affect user decision.
+**When to verify:** Before recommending batch sizes, connection pool settings, or schema designs
+where hitting a limit would cause failures. No need to verify for general guidance or when
+the exact number doesn't affect the user's decision.
 
-**Fallback:** If `awsknowledge` is unavailable, use the defaults above and flag that limits should be verified against [DSQL documentation](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/).
+**Fallback:** If `awsknowledge` is unavailable, use the defaults above and note to the user
+that limits should be verified against [DSQL documentation](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/).
```
anwesham-lab: why did we end up adding line breaks?

plugins/databases-on-aws/skills/dsql/SKILL.md
```diff
@@ -111,8 +111,8 @@ sampled in [.mcp.json](../../.mcp.json)
 
 ### Query Plan Explainability (modular):
 
-**When:** MUST load all four at Workflow 8 Phase 0 — [query-plan/plan-interpretation.md](references/query-plan/plan-interpretation.md), [query-plan/catalog-queries.md](references/query-plan/catalog-queries.md), [query-plan/guc-experiments.md](references/query-plan/guc-experiments.md), [query-plan/report-format.md](references/query-plan/report-format.md)
-**Contains:** DSQL node types + Node Duration math + estimation-error bands, pg_class/pg_stats/pg_indexes SQL + correlated-predicate verification, GUC experiment procedures + 30-second skip protocol, required report structure + element checklist + support request template
+**When:** MUST load [query-plan/workflow.md](references/query-plan/workflow.md) at Workflow 8 entry — it gates the remaining files
```
anwesham-lab: why remove the bullet structure?

plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/left-join-to-inner.md
```diff
@@ -0,0 +1,30 @@
+# Rewrite: LEFT JOIN with Null-Rejecting Predicate to INNER JOIN
+
+When a query uses LEFT JOIN but the WHERE clause rejects NULLs on the joined table, rewrite as INNER JOIN. This enables a simpler, more efficient join plan.
+
+**SHOULD apply when:** The WHERE clause rejects NULLs from the right-hand side of a LEFT JOIN (e.g., `IS NOT NULL`, equality comparisons, or any predicate that cannot be true for NULL).
+
+**Skip when:** NULLs from the right-hand side are intentionally preserved in the result.
```
anwesham-lab: is this an always? contextualize when/how often? Should this be an SHOULD Skip when?

plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/left-join-to-inner.md
```diff
@@ -0,0 +1,30 @@
+# Rewrite: LEFT JOIN with Null-Rejecting Predicate to INNER JOIN
+
+When a query uses LEFT JOIN but the WHERE clause rejects NULLs on the joined table, rewrite as INNER JOIN. This enables a simpler, more efficient join plan.
+
+**SHOULD apply when:** The WHERE clause rejects NULLs from the right-hand side of a LEFT JOIN (e.g., `IS NOT NULL`, equality comparisons, or any predicate that cannot be true for NULL).
```
anwesham-lab, line 5: is this a should or a must? should gives the model a directive to bypass when told something like to rush

plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/propagate-filter.md
```diff
@@ -0,0 +1,48 @@
+# Rewrite: Propagate Filter to JOIN Columns
```
anwesham-lab, line 1: since so many of these are deterministic, I question the meta structure of if we should instead leverage something like a python script converter of sorts that can parse and replace the SQL and the reference files just execute them?

anwesham-lab:
## PR #162 — Review Summary

`feat(dsql): enhance query plan explainability with type coercion detection, rewrites, and workflow extraction`

Reviewed at head SHA `07b6baaca029e3336ddfb03438eee26429734a72`. Sound direction; eval gains are real (especially eval 203 reltuples). **Holding** on five correctness bugs in the new SQL example pairs (rows 1–5) — agents will copy these into production rewrites that change result sets — plus a dangling cross-reference cluster (rows 6–8) where `workflow.md`, `plan-interpretation.md`, and `catalog-queries.md` reference an "implicit cast compatibility matrix" and a "Phase 5" that no longer exist after the rewrite. Structural, eval, and process items follow.

| # | Confidence | Area | Finding | Suggestion | Reviewed SHA |
|---|---|---|---|---|---|
| 1 | 95 | [query-rewrites/push-computation-to-constant.md L9-17](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/push-computation-to-constant.md#L9-L17) — correctness | First example is not equivalent under integer division. Original `WHERE emp_no * 100 / 5 = 10001` has no integer solution; rewrite `WHERE emp_no = 10001 * 5 / 100` matches `emp_no = 500`. The file's own "Skip when … integer-division rounding" caveat is violated by the leading example. | Replace with a genuinely invertible example (e.g. `emp_no + 100 = 10001` → `emp_no = 9901`), or use a numeric/float column. | 07b6baa |
| 2 | 90 | [query-rewrites/not-in-to-not-exists.md L1-26](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/not-in-to-not-exists.md#L1-L26) — correctness | "Sidesteps NULL semantics issues" understates: when the subquery contains NULL, NOT IN returns empty and NOT EXISTS returns the rows; the rewrite changes results, not just performance. | State explicitly: "NOT EXISTS does not preserve NOT IN's NULL-propagation; output differs when the subquery may contain NULLs. Confirm intent with the user before applying." | 07b6baa |
| 3 | 85 | [query-rewrites/subquery-unnesting-uncorrelated.md L9-23](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/subquery-unnesting-uncorrelated.md#L9-L23) — correctness | `SELECT DISTINCT R.*` collapses pre-existing duplicates in R that the original semi-join (`IN (SELECT …)`) preserved — fixes one duplicate problem by introducing another. | Either (a) recommend the EXISTS form (true semi-join) or (b) state the assumption "Apply only when S.b is unique (PK/UNIQUE); otherwise DISTINCT changes results." | 07b6baa |
| 4 | 85 | [query-rewrites/subquery-unnesting-correlated.md L20-26](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/subquery-unnesting-correlated.md#L20-L26) — correctness | Same DISTINCT-on-semi-join issue as #3 for the EXISTS→JOIN rewrite. | Same fix: prefer EXISTS, or document the uniqueness precondition. | 07b6baa |
| 5 | 85 | [query-rewrites/subquery-unnesting-scalar.md L33-52](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/subquery-unnesting-scalar.md#L33-L52) — correctness | `s_count` example: scalar `COUNT(*)` returns `0` for outer rows with no match; LEFT JOIN+GROUP BY rewrite returns NULL. Downstream `WHERE s_count = 0`, `SUM(s_count)`, etc. break silently. The first MAX example is fine (MAX returns NULL on empty). | Wrap with `COALESCE(Agg.s_count, 0) AS s_count`; add a one-line note that COUNT/SUM need COALESCE while MAX/MIN do not. | 07b6baa |
| 6 | 95 | [workflow.md L29-31](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md#L29-L31) — correctness | Trigger row references "Phase 5 re-entry for an existing report" but the workflow defines only Phase 0–4 (TOC L7–14). Routing L56 correctly says "append Addendum". | Replace with "Reassessment re-entry — re-runs Phase 1–2 and appends an Addendum per Phase 4." | 07b6baa |
| 7 | 90 | [plan-interpretation.md L194-242](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md#L194-L242), [workflow.md L100](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md#L100), [catalog-queries.md L122-124](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/catalog-queries.md#L122-L124) — correctness | Three files reference an "implicit cast compatibility matrix below/above/in plan-interpretation.md" that does not exist. The section was rewritten to recommend a live `pg_amop` query instead. Eval 201's expectation "Mentions implicit cast compatibility matrix" reinforces the phantom artifact. | Replace all four references with "the `pg_amop` query in catalog-queries.md (B-Tree Cross-Type Operator Support)." Update eval 201 expectation accordingly. | 07b6baa |
| 8 | 80 | [plan-interpretation.md L201-214](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md#L201-L214) — correctness | Bullet at L202 says "if an implicit cast exists, the planner can still use the index" — contradicts L211–214 which correctly notes B-Tree needs a registered cross-type *operator* (`pg_amop`), not just a `pg_cast`. The two paragraphs disagree. | Drop or rephrase L202: "If a cross-type B-Tree operator is registered (see pg_amop), the index can be used; otherwise the planner applies a per-row cast that defeats index ordering." | 07b6baa |
| 9 | 75 | [plan-interpretation.md L214](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md#L214) — correctness/durability | "Cross-type index support is limited to the integer family" stated as fact, no citation, no "verify before asserting" hedge. Will rot the moment DSQL adds a cross-type operator family. | Prefix with "At time of writing…" and route the agent through the `pg_amop` query before asserting this to a user. | 07b6baa |
| 10 | 70 | [catalog-queries.md L136](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/catalog-queries.md#L136) — correctness | `amopmethod = 10003` is a DSQL-internal magic number (PG mainline B-Tree is 403). No provenance comment; will silently break if the OID changes. | Add inline comment explaining provenance and a `SELECT oid FROM pg_am WHERE amname = 'btree'` recommendation as a hedge. | 07b6baa |
| 11 | 90 | [catalog-queries.md TOC L5-13](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/catalog-queries.md#L5-L13) — structure | TOC omits 3 of the 9 sections — all PR additions: "Column Types for Predicate Columns" (L107), "B-Tree Cross-Type Operator Support" (L125), "Indexed Column Types" (L157). | Add three TOC entries between current items 5 and 6. | 07b6baa |
| 12 | 90 | [plan-interpretation.md TOC L3-14](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md#L3-L14) — structure | TOC omits "Type Coercion and Index Bypass" (L186) — the headline new section of this PR. | Insert TOC entry, renumber subsequent items. | 07b6baa |
| 13 | 80 | [SKILL.md L112-115](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/SKILL.md#L112-L115) — structure | The PR rewrites this block but uses a single combined `When:`/`Contains:` while sibling "(modular):" sections give each sub-file its own `####` heading + per-file When/Contains. Loading conditions for `plan-interpretation.md`, `catalog-queries.md`, `guc-experiments.md`, `report-format.md`, and the rewrite indexes are no longer declared in the entry file. | Either give each query-plan reference its own `####` entry, or explicitly delegate routing to `workflow.md` and state that as the rule. | 07b6baa |
| 14 | 80 | [tools/evals/databases-on-aws/README.md](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/tools/evals/databases-on-aws/README.md) — multi-target sync | New `query_plan_rewrite_evals.json` is not added to the README's directory tree or per-tier eval section. Sibling evals (`evals.json`, `query_explainability_evals.json`) all have entries. The cluster-fixtures table also misses the new schemas (12-table join, 50M-row table). | Add the new eval and a fixtures row to the README. | 07b6baa |
| 15 | 75 | tools/evals/databases-on-aws/dsql/scripts/ — process | New eval JSON has no paired runner script under `scripts/`. Sibling evals all have one. PR ships only manual `query_plan_rewrite_eval_results.md`. | Either add `run_query_plan_rewrite_evals.py` (LLM-judge fits) or document explicitly that this suite is manual-only. | 07b6baa |
| 16 | 70 | [query_plan_rewrite_evals.json](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/tools/evals/databases-on-aws/dsql/query_plan_rewrite_evals.json) — tests | Coverage gap: 5 of 11 generic rewrites have no direct eval — `left-join-to-inner`, `propagate-filter`, `push-computation-to-constant`, `not-in-to-not-exists`, `flatten-union-all`. NOT IN→NOT EXISTS especially worth covering (correctness, not just perf). No negative cases (where the agent should *decline* the rewrite). | Add evals 206–212 covering missing patterns + at least one "OR across different columns → does NOT recommend OR-to-IN" negative case. | 07b6baa |
| 17 | 65 | [query_plan_rewrite_eval_results.md](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/tools/evals/databases-on-aws/dsql/query_plan_rewrite_eval_results.md) — tests | Sample size = 1 per cell, no model/version/temperature recorded, no variance analysis. PASS/FAIL is a single human transcript read. | Record model + version + n=3 with majority vote; add a `Runs` column; or downgrade the table to "qualitative comparison." | 07b6baa |
| 18 | 75 | [PR description](https://github.com/awslabs/agent-plugins/pull/162) — pr-body | PR body / commit `82617135` claim "275 lines (good)" but SKILL.md is 279 lines at head. Still under cap; cosmetic but it's a stated correctness claim. | Re-run `validate-size.py` on `07b6baa` and update the PR body / commit message. | 07b6baa |
| 19 | 65 | Multi-target sync (`awslabs/mcp`) — process | `awslabs/mcp@main` `src/aurora-dsql-mcp-server/skills/dsql-skill/references/query-plan/` does NOT contain `workflow.md`, `query-rewrites/`, or the new index files. PR description does not mention an MCP-mirror PR or follow-up. Per the dsql-skill-author placement rules, the default DSQL skill must propagate to the MCP standalone skill + Kiro Power. | Open a companion PR against `awslabs/mcp` mirroring the new files (and translate `workflow.md` for Kiro Power), or document explicitly that the mirror is out of scope and link the follow-up issue. | 07b6baa |
| 20 | 70 | [SKILL.md L172-173, L266-267](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/SKILL.md#L172-L173) — silent-failure | This PR softens the `awsknowledge` fallback rule from "flag that" to "note to the user that" — advisory phrasing, not a MUST. Agent can silently use stale defaults for decisions that turn on the exact value. | Promote to MUST: "MUST tell the user the lookup failed, MUST name the limit and value, MUST refuse the fallback when the recommendation depends on the exact value." | 07b6baa |
| 21 | 70 | [query-rewrites/reltuples-estimate.md](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/reltuples-estimate.md) + [eval 203](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/tools/evals/databases-on-aws/dsql/query_plan_rewrite_evals.json#L46-L56) — silent-failure | `reltuples` reflects last ANALYZE/autovacuum and may be drastically stale on a fresh or write-heavy table. Doc says "estimate, not exact" but does not require warning the user about staleness; eval 203 lacks the staleness expectation, so the failure mode is unobservable. | Add MUST: "Warn the user that `reltuples` reflects the last ANALYZE; recommend cross-checking `last_analyze` when the count drives a decision." Add eval expectation. | 07b6baa |
| 22 | 70 | [catalog-queries.md L107-180 (PR-added sections)](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/catalog-queries.md#L107-L180) — security | The 3 new sections this PR adds (`Column Types for Predicate Columns`, `B-Tree Cross-Type Operator Support`, `Indexed Column Types`) introduce fresh `'{schema}'` / `'{table}'` placeholder substitution patterns. SKILL.md Workflow 4 mandates `safe_query.build()` for query construction; these new examples teach lexical concatenation, an injection sink despite `readonly_query`. | Add a one-line MUST scoped to the new sections: "Substitute these placeholders via `safe_query.build()` with `ident()` — see input-validation.md." | 07b6baa |
| 23 | 60 | [query-rewrites/*.md (all 13)](https://github.com/awslabs/agent-plugins/tree/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites) — style | Every new rewrite file pairs `**SHOULD apply when:**` with `**Skip when:**`. The two are logical complements; per `authoring-style.md §Voice` reserve prohibition for irreversible harm. | Drop `Skip when:` and tighten `SHOULD apply when:`, or rephrase as a single `**Applies when:**` criterion. | 07b6baa |
| 24 | 80 | [workflow.md TOC L7-15](https://github.com/awslabs/agent-plugins/blob/07b6baaca029e3336ddfb03438eee26429734a72/plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md#L7-L15) — structure | TOC anchors encode the em dash with double hyphens (e.g., `#phase-0--load-reference-material`). GitHub collapses ` — ` to a single `-`, so all five Phase TOC links are broken in the rendered file. | Regenerate as `#phase-0-load-reference-material` … `#phase-4-produce-the-report-invite-reassessment` (and `#phase-3-experiment-conditional`). | 07b6baa |

**Reviewer scope.** This review covered the diff at the head SHA (21 files, +1131 / −36) — the new query-plan workflow extraction, type-coercion detection, 11-pattern rewrite library, and eval pair. Prior `amaksimo` review threads from the predecessor PR #161 (file split, RFC keywords, positive language, DATEADD→NOW()-INTERVAL, psql fallback removal) are addressed at this head; thank you.

---

> 🤖 This review was drafted with [Claude Code](https://claude.com/claude-code) using the `dsql-skill-author` Workflow 2 (reviewer) procedure and the 17+ sub-agent roster from `code-review.md`. Findings have been validated through the five-gate filter (re-read at head SHA, applicability, suggestion correctness, customer-value, confidence ≥ 60).
>
> **Was this review useful?** React with 👍 if the findings were helpful, 👎 if they missed the mark or introduced false positives. Reply with specifics so the review process can improve. Findings you disagree with are valid to push back on — confidence scores are not verdicts.

plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/propagate-filter.md
anwesham-lab: I think I still hold to this? Why are we relying on markdown applications alone? Why are these transformations not written into script patterns/tooling?

plugins/databases-on-aws/skills/dsql/SKILL.md
```diff
@@ -164,16 +166,17 @@ defaults that may change — when a user's decision depends on an exact limit, v
 | Max indexes per table          | 24            | `aurora dsql index limits`         |
 | Max columns per index          | 8             | `aurora dsql index limits`         |
 | IDENTITY/SEQUENCE CACHE values | 1 or >= 65536 | `aurora dsql sequence cache`       |
-| Supported column data types    | See docs      | `aurora dsql supported data types` |
 
-**When to verify:** Before recommending batch sizes, connection pool settings, or schema designs where hitting a limit would cause failures; any time the exact number can affect user decision.
+**When to verify:** Before recommending batch sizes, connection pool settings, or schema designs
```
anwesham-lab: > No need to verify for general guidance or when the exact number doesn't affect the user's decision.

rephrase to be prescriptive rather than avoidant prohibition?

plugins/databases-on-aws/skills/dsql/SKILL.md
```diff
@@ -164,16 +166,17 @@ defaults that may change — when a user's decision depends on an exact limit, v
...
+where hitting a limit would cause failures. No need to verify for general guidance or when
```
anwesham-lab: unnecessary and bad format/practice in skill linebreaks again? LOC consumption occupies additional token usage in skills frequently?

plugins/databases-on-aws/skills/dsql/SKILL.md
```diff
@@ -215,7 +218,10 @@ Every DDL statement generated in this workflow MUST be validated with `dsql_lint
 - MUST batch updates under 3,000 rows in separate transact calls
 - MUST issue each ALTER TABLE in its own transaction
 
-**Recovery — batch fails midway:** Rows already updated keep their new value (each batch committed independently). Resume by filtering on the unset state (`WHERE new_column IS NULL`) and continue. Re-running is safe because the filter naturally excludes completed rows.
+**Recovery — batch fails midway:** Rows already updated keep their new value (each batch committed
+in its own transaction). Resume by filtering on the unset state — e.g. add
+`WHERE new_column IS NULL` (or the sentinel value) to the next UPDATE — and continue from there.
+Re-running the entire migration is safe because the filter naturally excludes completed rows.
```
anwesham-lab: LOC bloating here again?

anwesham-lab:
## PR #162 — Multi-agent review (revised after empirical validation on a live DSQL cluster)

Reviewed at head SHA `1178334f37ce8abb22e6ee4955929ba1593d714e`. **Updated comment:** original revision had several findings that did not survive a per-finding §4 re-validation pass plus empirical testing on a live DSQL cluster (account 011528302527, region us-west-2, head-SHA-deployed test schema). The empirical results validate the PR's core thesis on every claim I tested: the type-coercion-bypass detection is real and load-bearing (cross-type `date = timestamp` was **85× slower** — Index Cond → Filter, 1.2ms → 102ms with 19,945 rows removed by post-scan filter), and the IN→EXISTS rewrite delivers a real ~**43× speedup** on DSQL (4,632ms Nested Loop → 108ms Hash Semi Join) because DSQL's planner does **not** auto-unnest `IN (SELECT…)` into a semi-join the way mainline PostgreSQL has since 8.4. Two original findings (`#9` projection drop, `#3` array/JSON regression depth) were graded false-positive / minor and have been removed; one (`#7`) is reframed because empirical evidence shows the hardcoded OID is correct on DSQL — only the verify-comment is misleading. Twenty review sub-agents ran in parallel, then one Opus grader per finding re-validated against head SHA. Findings below all survive at confidence ≥ 60 post-validation.

| # | Confidence | Area | Finding | Suggestion | Reviewed SHA |
|---|---|---|---|---|---|
| 1 | 90 | [SKILL.md L37-L40](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/SKILL.md#L37-L40) — correctness | `[.mcp.json](../../.mcp.json)` was rewritten to `[mcp/.mcp.json](mcp/.mcp.json)`. The new path resolves to `skills/dsql/mcp/.mcp.json`, which does not exist in the repo (verified via the GitHub contents API). The canonical file is at the plugin root, and `mcp/mcp-setup.md` confirms ".mcp.json at the plugin root". | Revert to `[.mcp.json](../../.mcp.json)`. | `1178334f` |
| 2 | 90 | [SKILL.md L177-L180](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/SKILL.md#L177-L180) — correctness | `[scripts/](../../scripts/)` and `[scripts/README.md](../../scripts/README.md)` were rewritten to `scripts/...` — resolves under the skill, where no `scripts/` subdirectory exists. The trailing "and hook configuration" wording was also dropped. `validate-references.py` does not flag these because the new paths lack its trigger keywords. | Revert both paths to `../../scripts/...` and restore the dropped wording. | `1178334f` |
| 3 | 75 | [SKILL.md awsknowledge limits table](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/SKILL.md#L160-L172) — regression | Deletion of `\| Supported column data types \| See docs \| aurora dsql supported data types \|`. PR #155 added this row specifically "so the skill does not drift as DSQL's type surface evolves." Removed without replacement and out of stated PR scope. | Restore the row, or call out the deletion in the PR body with rationale. | `1178334f` |
| 4 | 90 | [`subquery-unnesting-scalar.md` L5](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/subquery-unnesting-scalar.md#L5) — correctness | `For COUNT and SUM, MUST wrap with COALESCE(..., 0) because the LEFT JOIN returns NULL (not 0) for unmatched rows — the scalar subquery returns 0.` This is wrong for `SUM`. Per the [PostgreSQL aggregate-functions docs](https://www.postgresql.org/docs/16/functions-aggregate.html): *"Most aggregate functions, except `count`, return null when no rows are selected. For example, `sum` of no rows returns null, not zero."* Both the scalar subquery and the LEFT-JOIN form return NULL for `SUM` over empty sets — only `COUNT` differs. Wrapping `SUM` with `COALESCE(..., 0)` silently changes results. | Restrict the COALESCE rule to `COUNT` only. For `SUM`/`MIN`/`MAX`, omit COALESCE so the rewrite preserves NULL semantics. | `1178334f` |
| 5 | 70 | [`catalog-queries.md` L141, L158](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/catalog-queries.md#L138-L160) — wording / verifier-misleading | (Original finding reframed after empirical validation.) The hardcoded `WHERE ao.amopmethod = 10003` is **correct** on DSQL — every DSQL index uses the `btree_index` access method (OID 10003); the `403/btree` access method exists in `pg_am` but is not used by any actual index. The empirical pg_amop check at 10003 correctly excludes `date<->timestamp` (which produced an 85× slowdown via post-scan filter on a 20k-row test table), while at 403 it incorrectly includes it. **However**, the inline verify-comment `Verify with: SELECT oid FROM pg_am WHERE amname = 'btree'` would mislead a verifier — that returns 403 (regular btree, `amtype='i'`), not 10003 (`btree_index`). | Update the verify-comment to `Verify with: SELECT oid FROM pg_am WHERE amname = 'btree_index'` (or, equivalently, switch the query to `WHERE ao.amopmethod = (SELECT oid FROM pg_am WHERE amname = 'btree_index')` so the OID lookup is the database's responsibility). | `1178334f` |
| 6 | 78 | [`catalog-queries.md` L109-L111](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/catalog-queries.md#L109-L111) — correctness | The `MUST substitute … via safe_query.build() with ident()` directive is scoped to one section, but the same `'{schema}'`/`'{table}'`/`'{col}'` placeholders are used throughout the file. Worse, per [`mcp/tools/input-validation.md`](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/mcp/tools/input-validation.md), `ident()` emits `"value"` (double-quoted identifier) and is for table/column names. The placeholders here sit in single-quoted string-literal positions (`c.table_schema = '{schema}'`, `n.nspname = '{schema}'`, `IN ('{table1}', '{table2}')`) which need `regex()`/`allow()` — both emit `'value'`. Following the directive verbatim produces invalid SQL like `WHERE c.table_schema = "public"`. (Note: positions like `FROM {schema}.{table}` and `GROUP BY {column}` legitimately need `ident()`.) | Lift the substitution rule to the file preamble with per-position guidance: identifier positions (`{schema}.{table}`, `GROUP BY {column}`) → `ident()`; literal positions (`= '{schema}'`, `IN ('{table}')`) → `regex()` or `allow()`. | `1178334f` |
| 7 | 70 | [`workflow.md` L42-L46](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md#L42-L46) — correctness regression | The `Context Disambiguation` row says "offer the psql fallback" when no MCP is connected, but `workflow.md` never reproduces the fallback (auth-token command, `psql <<<` heredoc, `$?` check) the prior SKILL.md Workflow 8 carried. [Phase 1 L78](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md#L76-L80) and [Safety L127](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md#L120-L130) say `readonly_query` exclusively, contradicting the offered fallback. | Either restore the psql fallback procedure, or change the row to "MUST refuse — no MCP connection means no plan capture." | `1178334f` |
| 8 | 70 | [`split-large-joins.md`](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/split-large-joins.md#L1-L43) — example consistency | File states "e.g., 10 joins for Aurora DSQL" but the worked Original example only joins 7 tables (R1–R7), so by its own rule the rewrite would not trigger. Eval 202 uses 12 tables (consistent with a >10 threshold) — the example is the outlier. The final `JOIN sub2 ON sub1.id = sub2.id` is also ambiguous: `SELECT *` from each CTE propagates multiple `id` columns. | Bump example to >10 tables and project explicit columns / alias the join keys per CTE. | `1178334f` |
| 9 | 80 | [`plan-interpretation.md` L227-L233](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/plan-interpretation.md#L227-L233) — example correctness | Recommendation Template shows `WHERE col = '42'` rewritten as `WHERE col = 42::float` regardless of column type. The doc's own type matrix says cross-type B-Tree support is integer-family-only — `::float` is the wrong example type entirely. Eval 201's `expected_output` uses `'12345'::integer`, which is consistent with the matrix. | Re-anchor the example to the column's actual type (e.g., for `bigint` columns: `col = 42` or `col = 42::bigint`), and align with eval 201's `::integer` casting. | `1178334f` |
| 10 | 65 | [`workflow.md` Phase 0 L60-L74](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/workflow.md#L60-L74) vs [SKILL.md L113-L117](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/SKILL.md#L113-L117) — accuracy | SKILL.md describes `workflow.md` as "follow its loading instructions rather than loading all files upfront", but Phase 0 mandates `MUST read these four files before starting` plus `SHOULD also load` two indexes — six files at workflow entry, the opposite of lazy loading. (The 13 individual rewrite sub-files are lazy-loaded — that's the genuine lazy-load surface.) | Reword SKILL.md's `**Contains:**` to drop the misleading "rather than loading all files upfront" claim; describe the actual loading model ("loads 4–6 files at Phase 0; rewrite sub-files on-demand at Phase 2"). | `1178334f` |
| 11 | 75 | [`in-subquery-to-exists.md`](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/in-subquery-to-exists.md) + [`subquery-unnesting-uncorrelated.md`](https://github.com/awslabs/agent-plugins/blob/1178334f37ce8abb22e6ee4955929ba1593d714e/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/subquery-unnesting-uncorrelated.md) — redundancy / routing | Both files target the same `IN (SELECT …)` → `EXISTS` rewrite. Empirically validated as valuable on DSQL (43× speedup vs. mainline PG's auto-unnesting). The duplication remains: structurally identical examples, and Eval 200's `expected_output` cites them interchangeably ("subquery-unnesting-uncorrelated.md or in-subquery-to-exists.md") — a tell that the routing is undecidable. The "Uncorrelated IN-subquery" / "Large IN-subquery result set" split in `query-rewrites-generic.md` carries no deterministic predicate the agent can evaluate at trigger-time. | Merge into one canonical file (`subquery-unnesting-uncorrelated.md` is the broader one — it covers the JOIN alternative and correlation gate; recommend deleting `in-subquery-to-exists.md` and routing all uncorrelated-IN-to-EXISTS prompts there). Keep correlated and scalar variants separate (genuinely different input shapes). | `1178334f` |

Findings dropped after re-validation (full transcripts available on request):

- **(dropped, false positive)** `not-in-to-not-exists.md` "Additional example" projection drop — re-read of L29-L48 shows Original is `SELECT product_id FROM products`, not `SELECT *`; projection is preserved (with table-qualification only).
- **(dropped, < 60)** Array/JSON storage rule revert — SKILL.md routes to `development-guide.md` as authoritative, dev-guide still carries the longer correct form, so SKILL.md is a stale shorthand rather than a behavior regression.
- **(dropped, < 60)** Eval results 206–210 missing transcripts — real but borderline; matches the PR body's explicit "n=1, manual qualitative" framing.
- **(dropped, < 60)** Manual-only eval suite has no automated grader — author has explicitly committed to a future Python converter; tracking-debt rather than blocking.
- **(dropped, < 60)** Three unresolved reviewer threads — process item, not a code defect.

Empirical context (run on live DSQL cluster):

- **Type-coercion bypass detection (the PR's headline feature) is real and load-bearing.** A 20,000-row `test_typecoerce(d date) USING btree_index` showed `WHERE d = '2024-06-01'::date` → Index Cond, 1.2ms; `WHERE d = '2024-06-01'::timestamp` → Filter, 19,945 rows removed by Filter, 102ms. The cross-type pair is empirically present in `pg_amop @ amopmethod=403` and absent at `amopmethod=10003`, exactly matching the skill's detection logic.
- **IN→EXISTS rewrite delivers ~43× speedup on DSQL.** `WHERE customer_id IN (SELECT customer_id FROM orders WHERE order_date > '2024-06-01')` → Nested Loop, 4,632ms; equivalent EXISTS form → Hash Semi Join, 108ms. Mainline PG ≥ 8.4 auto-unnests both into a semi-join; DSQL does not. The rewrite is genuinely valuable on DSQL — author was right.
- **Propagate-filter rewrite delivers a cardinality-estimate improvement** but no significant timing delta at this scale (84ms vs. 81ms). The planner does not derive the transitive predicate automatically. Effect would compound on larger joins.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anwesham-lab:
I think it's worth using our self-review skill and doing a couple of explicit passes with subagents deployed from the code-review and pr-toolkit-review plugins to get to an explicit convergence state that you can audit and post to log changes being made.

amaksimo:
> Extract Workflow 8 from SKILL.md into references/query-plan/workflow.md (SKILL.md: 334 → 281 LOC)


I do not see this? Was it accidentally left out of the PR?


The Workflow 9 section currently only links the four Phase-0 files, nothing reaches workflow.md or the rewrites, so validate-references flags all 16 new query-plan files as orphans and the agent never loads them. Curious how the evals work then?

We can add to the Workflow 9 section to query-plan/workflow.md to make all 16 reachable (the rewrite indexes chain from there).

plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/subquery-unnesting-scalar.md
```diff
@@ -0,0 +1,61 @@
+# Rewrite: Subquery Unnesting — Scalar
+
+When a query contains a scalar subquery in the SELECT clause computing an aggregate correlated by equality, rewrite it as a LEFT JOIN with GROUP BY. This reduces repeated subquery executions and enables better join planning.
+
+**SHOULD apply when:** The scalar subquery is correlated via equality and contains an aggregate function (MAX, MIN, COUNT, SUM). For COUNT and SUM, MUST wrap with `COALESCE(..., 0)` because the LEFT JOIN returns NULL (not 0) for unmatched rows — the scalar subquery returns 0.
```
amaksimo: I think there is a mistake here, the examples look right, but the wording here should probably be:
```
For COUNT, MUST wrap with COALESCE(..., 0) because COUNT returns 0 for an empty input while LEFT JOIN produces NULL for missing groups. Other aggregates (SUM, MIN, MAX, AVG) MUST NOT be wrapped with COALESCE because scalar aggregate subqueries already return NULL for empty inputs.
```

What is written is backwards for SUM. Eg. (SELECT SUM(y) FROM S WHERE S.x = R.x) returns NULL over no matching rows which is same as the LEFT JOIN thus wrapping it in COALESCE(..., 0) changes the result (NULL→0).

Only COUNT returns 0 for an empty set. 


Tested like:

SUM over no rows  : None     # NULL matches LEFT JOIN, no COALESCE needed
COUNT over no rows: 0        # COALESCE needed
MAX over no rows  : None


So MUST COALESCE should apply to COUNT only.... SUM/MIN/MAX/AVG must not be coalesced or the rewrite diverges from the original.

plugins/databases-on-aws/hooks/hooks.json
amaksimo, line 5: this branch is on the transact hook, but the workflow requires EXPLAIN to run via readonly_query (transact is explicitly forbidden for plan capture).... so it'll never fire. either add a readonly_query matcher or drop it. (also missing a period at the end.)


btw, we should re-run evals after this, I would expect different results.

anwesham-lab: iirc, James explicitly encountered this path resulting in the addition AND, technically transact is frequently used to bundle multiple read queries in a single transaction by the AI, so I'm not opposed to the inclusion here.

anwesham-lab:
## PR #162 — Review (merge-blocking findings only)

Reviewed at head SHA `4523be94bc10f2e07c64243f57d7b9d82d8880db` via a 20+ sub-agent parallel pass, with DSQL claims validated empirically against a live Aurora DSQL cluster (us-west-2, PostgreSQL-16-compatible). The PR's core thesis holds up: the type-coercion index-bypass detection is **real and load-bearing** (cross-type `date = timestamp` literal demoted the predicate from `Index Cond` to `Filter`, `Rows Removed by Filter: 5996`, ~1.2 ms → ~33 ms on a 6k-row table), `amopmethod = 10003` is the correct DSQL B-Tree method (`btree_index`=10003 vs mainline `btree`=403), the "integer-family-only" cross-type claim is accurate, and the COUNT-only-COALESCE wording in `subquery-unnesting-scalar.md` is correct. Those are confirmed, not findings.

Five **merge-blocking** items remain — three live-cluster-validated correctness defects in content **introduced by this PR**, plus two content/PR-body accuracy defects that ship a wrong fact into the agent context. Non-blocking style/structure items are omitted by request.

| # | Confidence | Area | Finding | Suggestion | Validation (how confirmed) | Reviewed SHA |
|---|---|---|---|---|---|---|
| 1 | 90 | [catalog-queries.md L5-L8](https://github.com/awslabs/agent-plugins/blob/4523be94bc10f2e07c64243f57d7b9d82d8880db/plugins/databases-on-aws/skills/dsql/references/query-plan/catalog-queries.md#L5-L8) — correctness / security | The new substitution rule pairs quoted placeholders (`= '{schema}'`, `IN ('{table}')`) with the `allow()`/`regex()` helpers, which **emit their own single quotes** (`'value'`). Following the rule verbatim produces `WHERE n.nspname = ''public''` — an empty string adjacent to a bare token. Every string-literal template in the file (e.g. L57-59, L82-83, L186-187) inherits this. | Strip the literal quotes from string-literal placeholders so the helper supplies them: `= {schema}` with `schema=regex(...)`, `IN ({table1}, {table2})`. Add one worked `safe_query.build()` example. (`ident()` for identifier positions likewise must not be re-quoted.) | Reproduced on the live cluster: `EXPLAIN SELECT 1 WHERE 'a' = ''public'';` → `ERROR: syntax error at or near "public"`. Confirmed `regex()`/`allow()` emit `'value'` per `safe_query.py` self-test. | `4523be94` |
| 2 | 90 | [split-large-joins.md L3](https://github.com/awslabs/agent-plugins/blob/4523be94bc10f2e07c64243f57d7b9d82d8880db/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/split-large-joins.md#L2-L4) (also example at [L12](https://github.com/awslabs/agent-plugins/blob/4523be94bc10f2e07c64243f57d7b9d82d8880db/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/split-large-joins.md#L11-L13)) — correctness | States "the optimizer's DP threshold (e.g., **10 joins** for Aurora DSQL)" and the worked example says "11 tables — exceeds DP threshold of 10." The actual DSQL value is **8**, so the stated fact is wrong and the example splits at the wrong boundary. | Change to 8 (`join_collapse_limit`/`from_collapse_limit`); fix the example accordingly. Better: instruct the agent to `SHOW join_collapse_limit;` on the target cluster rather than hardcoding, since the file already (correctly) names those GUCs. Optionally note `geqo_threshold=12` as the DP→GEQO crossover. | Live cluster: `SHOW join_collapse_limit;` → **8**, `SHOW from_collapse_limit;` → **8**, `SHOW geqo_threshold;` → 12. Neither matches "10." | `4523be94` |
| 3 | 90 | [reltuples-estimate.md L7](https://github.com/awslabs/agent-plugins/blob/4523be94bc10f2e07c64243f57d7b9d82d8880db/plugins/databases-on-aws/skills/dsql/references/query-plan/query-rewrites/reltuples-estimate.md#L6-L8) — correctness | The staleness-mitigation step is the file's main safeguard, but it recommends cross-checking `pg_stat_user_tables.last_analyze` — a view that is **never populated on DSQL** (no PG stats collector / autovacuum). The agent is told to perform an impossible check, so the staleness guard silently does nothing. Separately, `reltuples` is `-1` (the PG "unknown" sentinel) on a never-populated table, so `reltuples::bigint AS estimated_count` (L17) can return `-1`. | Drop or replace the `pg_stat_user_tables.last_analyze` cross-check with DSQL-appropriate staleness guidance. Guard the estimate with `GREATEST(reltuples, 0)::bigint` (or treat negative as "unknown"). | Live cluster: `SELECT count(*) FROM pg_stat_user_tables;` → **0 rows**; `last_analyze` empty even after manual `ANALYZE`. A freshly created table reported `reltuples = -1`. | `4523be94` |
| 4 | 90 | [SKILL.md L87](https://github.com/awslabs/agent-plugins/blob/4523be94bc10f2e07c64243f57d7b9d82d8880db/plugins/databases-on-aws/skills/dsql/SKILL.md#L86-L88) — correctness / content | The entry file advertises "Index of **11** generic rewrite patterns," but the generic index lists **10** (L9-L18 of `query-rewrites-generic.md`) and only 10 generic files exist (12 total = 10 generic + 2 DSQL-specific). This wrong count ships into every agent context that loads SKILL.md and can send the agent hunting for a non-existent 11th pattern. The PR description repeats the same "11 generic" error. | Change "11" to "10" at SKILL.md L87 (and in the PR body), or add the missing 11th pattern if one was intended. Prefer dropping the hardcoded count to avoid future drift. | Counted directly at head SHA: `query-rewrites-generic.md` table = 10 rows; `ls query-rewrites/*.md` = 12 files (10 generic + reltuples-estimate + split-large-joins). | `4523be94` |
| 5 | 95 | [PR description](https://github.com/awslabs/agent-plugins/pull/162) — pr-body | The Summary claims "SKILL.md: 334 → 281 LOC" and Validation claims "`validate-size.py`: 281 lines." Both numbers are wrong: SKILL.md is **249 lines** at head (base was 246), so it **grew by 3**, not shrank to 281. The stated extraction-driven reduction never happened in this PR (the phase procedures already lived in the four reference files on `main`; this PR adds the new `workflow.md` orchestration layer and rewires links). The false numbers already triggered a maintainer's "I do not see this?" question. | Correct the PR body to "SKILL.md: 246 → 249 LOC" and "validate-size.py: 249 lines (good)," and reframe the bullet as wiring references to the new `workflow.md` rather than a 53-line extraction. | Authoritative: `wc -l` = 249 at head, `git show main:…SKILL.md \| wc -l` = 246; ran `tools/validate-size.py` → reports **249 lines / good**, not 281. | `4523be94` |

Items 1-3 reproduce on a real cluster and are introduced by this PR's new files (1 and 3 are silent-failure-class — wrong-but-runs / does-nothing; 2 is a wrong hardcoded fact). Items 4-5 ship an incorrect fact into the agent context (4) and misrepresent the change in the PR record (5). Recommend addressing all five before merge.

> Validated through the five-gate filter (re-read at head SHA, applicability confirmed to this diff, suggestion checked, customer-value, confidence ≥ 60 post-validation). DSQL behavior empirically verified on a live cluster, not asserted from training data.

<sub>If this review was useful, react 👍; otherwise 👎. Confidence scores are not verdicts — push back on anything you disagree with.</sub>

anwesham-lab (dismissed):

krokoko (changes requested):
Please bump the plugin version in the required files, thanks !

Morlej:
> Please bump the plugin version in the required files, thanks !

Done!

anwesham-lab (dismissed):

krokoko (approved):

anwesham-lab (approved):

Issue #163, opened 2026-05-12
RFC: aws-transform plugin


youtuyy:
### Is this related to an existing feature request or issue?

Related MCP server contribution: [awslabs/mcp#2811](https://github.com/awslabs/mcp/issues/2811) (RFC: AWS Transform MCP Server), landed via [awslabs/mcp#3282](https://github.com/awslabs/mcp/pull/3282) as `src/aws-transform-mcp-server/`. This RFC proposes the companion plugin that brings AWS Transform's domain expertise to coding agents and composes with that MCP.

### Summary

Add an `aws-transform` plugin for migration architects, modernization leads, and partner consultants — users who arrive with a workload to migrate rather than an AWS API to call. AWS Transform is AWS's AI-powered code and workload modernization service, and the plugin wraps it into a focused, persona-scoped tool surface (assess, plan, transform, validate) backed by the [`awslabs.aws-transform-mcp-server`](https://github.com/awslabs/mcp/tree/main/src/aws-transform-mcp-server) (published to PyPI, maintained by the AWS Transform service team).

AWS Transform launched a Kiro Power on 2026-04-30 bundling its MCP and skills. This RFC and the accompanying PR deliver the equivalent experience to users of Claude Code, Cursor, Codex, and other agents that install from `awslabs/agent-plugins`, targeting a 2026-05-14 launch alongside the broader AWS Transform anniversary release.

### Use case

The plugin serves three concrete persona use cases:

- **Migration architects** planning and executing workload moves to AWS:
  - .NET Framework → .NET 8/10 on Linux (project file migration, API replacements, runtime compatibility)
  - Mainframe COBOL → Java (program decomposition, dependency analysis, assessment)
  - VMware → EC2 (discovery via connector, landing zone setup, network migration, server rehost via MGN, containerization to ECS/EKS)
  - SQL Server / Oracle / MySQL → Amazon Aurora (schema conversion, data migration planning)
- **Modernization leads** driving in-place code upgrades:
  - Java 8/11/17 → 17/21
  - Python 2 → 3
  - Node.js version bumps
  - AWS SDK v2 → v3 migrations
- **Partner consultants** running user-defined transformations across customer fleets — single-repo local execution for fast iteration, or many-repo remote execution for scale.

The plugin can grow within the migration-and-modernization persona over time by absorbing related tooling (for example, DMS, MGN, Migration Hub, Application Discovery Service) under one installable rather than splintering into per-service plugins.

#### Structure

```
[...]

PR #164, opened 2026-05-12
feat(aws-transform): add AWS Transform plugin

youtuyy:
## Summary

Adds the **aws-transform** plugin, which brings [AWS Transform](https://aws.amazon.com/transform/) — AWS's AI-powered code and workload modernization service — into coding agents.

Covers the full AWS Transform product surface:
- **.NET Framework → .NET 8/10** (or latest LTS), Linux-compatible
- **Mainframe COBOL → Java** with assessment and decomposition
- **VMware → EC2** (discovery, landing zone, network migration, server rehost via MGN, containerization)
- **SQL Server / Oracle / MySQL → Amazon Aurora** (schema conversion and migration planning)
- **Language upgrades**: Java 8/11/17 → 17/21, Python 2 → 3, Node.js bumps, **AWS SDK v2/v3 migrations**
- **Custom transformations**: run user-defined transformation definitions across single repos (local mode) or many in parallel (remote mode)

The plugin depends on the existing [`awslabs.aws-transform-mcp-server`](https://github.com/awslabs/mcp/tree/main/src/aws-transform-mcp-server) (published to PyPI) and complements it with domain guidance — workflow structure, workload-specific decision points, authentication handling, and troubleshooting.

## RFC

Linked to [RFC #163](https://github.com/awslabs/agent-plugins/issues/163). Mirrors the parallel approach taken for the AWS Transform MCP server (RFC [awslabs/mcp#2811](https://github.com/awslabs/mcp/issues/2811) + PR [awslabs/mcp#3282](https://github.com/awslabs/mcp/pull/3282)).

## Changes

- **New plugin**: `plugins/aws-transform/`
  - `.claude-plugin/plugin.json` and `.codex-plugin/plugin.json` (schema-validated, Migration category)
  - `.mcp.json` referencing the PyPI-published MCP server
  - Single `aws-transform` skill with workload-scoped references (auth, tools, workflow, .NET, mainframe, SQL, VMware subtree, custom-transformation subtree)
- **Marketplace registration**:
  - `.claude-plugin/marketplace.json` — added aws-transform entry under `migration` category
  - `.agents/plugins/marketplace.json` — added aws-transform entry under `Migration` category
- **README**: added row to the Plugins table

## Checklist

- [x] `.claude-plugin/plugin.json` validates against `schemas/plugin.schema.json`
- [x] `.codex-plugin/plugin.json` validates against `schemas/codex-plugin.schema.json` (including required `interface` fields)
- [x] `.mcp.json` validates against `schemas/mcp.schema.json`
- [x] `SKILL.md` frontmatter validates against `schemas/skill-frontmatter.schema.json`
- [x] Plugin registered in both marketplace files with consistent category
- [x] README plugins table updated

---


plugins/aws-transform/test-prompts.json
linchenk369, line 1: We need to drop this. This is for convertor skill validation.

plugins/aws-transform/README.md
```diff
@@ -0,0 +1,94 @@
+# AWS Transform Agent Plugin
+
+Migrate and modernize legacy codebases to AWS. Covers .NET Framework to .NET 8/10, mainframe COBOL to Java, VMware to EC2, SQL Server to Aurora PostgreSQL, and Java/Python/Node.js version upgrades and AWS SDK migrations.
+
+Works with Claude Code, Kiro CLI, and Codex.
+
+## Quick Install
```
linchenk369: We need to make this external compatible. We should not have cloudfront link here

plugins/aws-transform/test-prompts.json
youtuyy: Removed — test-prompts.json was only used by the internal converter skill's validation script; no public use. Dropped in 9312462.

plugins/aws-transform/README.md
youtuyy: Rewrote the README to match the marketplace-install pattern used by other plugins (`/plugin marketplace add awslabs/agent-plugins` + `/plugin install aws-transform@agent-plugins-for-aws`). Dropped the `<distribution>` placeholders and `install.sh` references entirely — those were for the internal CloudFront-distributed zip, which doesn't apply here. Also added the missing CODEOWNERS entry for plugins/aws-transform.

krokoko:
Automated review:

1. Blanket silent error suppression on resume [silent-failure-hunter]

  Location: SKILL.md (resume section) + references/workflow.md

  ▎ "If any call fails for any reason, silently fall back to local context. Do NOT reveal your reasoning about the refresh to the
  ▎  user"

  This catch-all suppression swallows network outages, expired credentials, revoked access, and service errors. A user could
  resume on stale local context (assessment says "passed"), build requirements, then discover the real state was "failed" only
  after investing significant effort.

  Recommendation: Distinguish between "auth not configured" (fine to skip silently) and "auth configured but call fails" (user
  must be informed with actionable message like "Could not reach the service — continuing with last known state. Sign in to
  refresh.").

  2. Contradictions between SKILL.md constraints and reference files [comment-analyzer]

  - Agent hardcoding: SKILL.md says "MUST discover agents dynamically — do not hardcode agent names." But dotnet.md and
  workflow.md explicitly hardcode dotnet-chatty-agent. An agent following SKILL.md's absolute rule would refuse what the .NET
  reference demands.
  - Telemetry contradiction: custom.md Greeting includes "Note: this skill collects telemetry" but the Telemetry section says
  "NEVER mention telemetry to the user." An AI agent cannot follow both simultaneously.

  Recommendation: Add an explicit exception in SKILL.md for the .NET agent. For telemetry, reconcile the two rules (e.g., "except
   for the initial greeting disclosure").

  ---
  Important Issues (6 found)

  3. create_job silently defaults to wrong config on malformed JSON [silent-failure-hunter]

  Location: references/dotnet.md

  ▎ "The backend parses objective as JSON — if it receives prose, parsing fails silently and defaults to auto + net8.0."

  User selects .NET 10 interactive, but malformed JSON silently overrides to .NET 8 auto. No verification step is prescribed
  after job creation.

  Recommendation: Add instruction to verify effective configuration via get_resource(resource="job") after creation.

  4. upload_artifact silently succeeds but is functionally broken without planStepId [silent-failure-hunter]

  Location: references/vmware-network.md

  File uploads without planStepId succeed (HTTP 200) but are invisible to the agent. Also ~/file.zip "fails silently." The doc
  flags this as "CRITICAL" but only as a reminder — no verification step is prescribed.

  Recommendation: Instruct agent to verify upload by listing plan step artifacts immediately after.

  5. MCP server startup failure has no guidance [silent-failure-hunter]

  Location: .mcp.json, SKILL.md

  If uvx is missing or awslabs.aws-transform-mcp-server@latest fails to download, the MCP server silently fails to start. No
  instructions in SKILL.md for this scenario.

  Recommendation: Add guidance to check uv --version and provide installation instructions if tools are unavailable.

  6. load_instructions and configure tools undocumented [comment-analyzer]

  Location: references/dotnet.md references these tools, but tools.md (the tool reference) doesn't mention them.

  load_instructions "gates all job-scoped tools — MUST be called once per job" — this is critical to document.

  7. adaptive_poll tool referenced but undefined [comment-analyzer]

  Location: references/workflow.md

  The text says "use adaptive_poll" but this tool is not documented anywhere in the PR. Implementers won't know what it is.

  8. Workflow phase names don't align with context.json schema [comment-analyzer]

  SKILL.md lists 9 phases (Resume, Intent, Discovery, Scope, Assessment, Requirements, Approval, Tasks, Execute). The
  context.json phase field in workflow.md uses 8 different values
  (intent|discovery|scoped|assessed|requirements|planning|executing|complete). No mapping is provided.

linchenk369:
Thanks for the thorough review! Here's our assessment of each issue:

### Addressed

**Issue 2a — Agent discovery vs hardcoded .NET agent name:**
The SKILL.md constraint says "do not hardcode agent names," but `dotnet.md` explicitly overrides this with `dotnet-chatty-agent`. This is intentional — the .NET agent is a special case, and the workload reference is authoritative for agent selection within its domain.

**Issue 2b — Telemetry greeting contradiction:**
`custom.md` greeting discloses telemetry to the user, while the Telemetry section says "NEVER mention telemetry to the user." Direct contradiction within the same file. We're reaching out to the custom workload team to address this.

**Issue 8 — Phase name misalignment:**
The SKILL.md workflow lists 9 named steps but `context.json` schema has 8 phase values with no explicit mapping between them. We'll align the two.

### Post-launch quick fix

**Issue 3 — Malformed JSON defaults silently:**
`dotnet.md` documents that if `objective` receives prose instead of valid JSON, parsing fails silently and defaults to `auto` + `net8.0` — but prescribes no verification step. We'll add a post-`create_job` verification note to confirm the job is using the user's chosen settings. Tracking as a post-launch quick fix.

### By design / False positive

**Issue 1 — Silent error suppression in resume:**
This is intentional UX. The resume flow deliberately distinguishes between auth-not-configured (skip refresh entirely) and service errors (silently fall back to local context). The goal is to never block the user or expose internal mechanics — the user sees only the resume message. The error handling is explicit in the doc, just silent to the end user.

**Issue 4 — Broken artifact uploads without `planStepId`:**
False positive. The MCP schema for `upload_artifact` defines `planStepId` as optional with a `null` default. Uploads work correctly without it.

**Issue 5 — MCP server startup failure undocumented:**
The MCP server is installed via `uvx` using standard Python packaging. Troubleshooting `uvx`/Python installation issues is outside the scope of this skill — it's a general environment concern, not specific to AWS Transform.

**Issues 6 & 7 — Missing docs for `load_instructions`, `configure`, `adaptive_poll`:**
False positive. `tools.md` opens with: *"Canonical tool names, parameters, and per-resource requirements come from the MCP server's `tools/list` response — read the tool descriptions directly."* Individual MCP tools are intentionally not documented in tools.md — the MCP server is the source of truth for tool definitions. tools.md only covers cross-tool workflows, behavioral rules, and CLI commands that the MCP schemas can't describe.

### Note on workload references

We don't control the `workload-*.md` files (dotnet, vmware, mainframe, sql, custom) — those are authored and maintained by the respective workload teams. Issues 2b and 3 fall within those files, so we're coordinating with the relevant teams.

linchenk369:
Follow-up on a few items:

**Issue 2a — Agent discovery vs hardcoded .NET agent name:**
The .NET workload team confirmed this is expected behavior for release. The `dotnet-chatty-agent` hardcode in `dotnet.md` is an intentional override — the .NET agent is a special case where dynamic discovery is not needed. The workload team will address the SKILL.md constraint wording as a post-launch fix.

**Issue 8 — Phase name misalignment:**
After inspection, the mapping is sound. The 9 workflow steps and 8 context.json phase values align correctly:
- "Resume" (Step 1) is the entry point — not a persistable phase.
- "Approval" (Step 7) is a gate within the requirements→planning transition, not a distinct phase.
- `complete` is a terminal state in the schema with no corresponding workflow step (the workflow ends at Execute).
- "Tasks" (Step 8) maps to `planning` — the phase name reflects the broader activity, not just task generation.

No changes needed here.

linchenk369 (dismissed):

krokoko:
Missing INTERFACE_METADATA in Codex manifest generator — build will break

  [code-reviewer] tools/generate_codex_manifests.py

  The repo auto-generates .codex-plugin/plugin.json and .agents/plugins/marketplace.json from the Claude marketplace. This PR hand-crafts both, but doesn't add an INTERFACE_METADATA entry in the generator
  script. On the next mise run build, the script will raise ValueError("Missing interface metadata...") because aws-transform exists in .claude-plugin/marketplace.json but has no matching entry. The
  hand-crafted Codex manifest will also be overwritten.

  Fix: Add an aws-transform entry to INTERFACE_METADATA in tools/generate_codex_manifests.py.

  2. Hardcoded .NET agent name contradicts "discover dynamically" constraint

  [comment-analyzer] SKILL.md vs dotnet.md / workflow.md

  SKILL.md mandates: "MUST discover agents dynamically via list_resources — do not hardcode agent names." But workflow.md explicitly says the .NET agent is hardcoded, and dotnet.md uses dotnet-chatty-agent as
  a literal. No exception clause exists in SKILL.md.

  Fix: Add an explicit exception in SKILL.md's constraints, or make .NET discovery-based like other workloads.

  3. Inconsistent MCP tool names in sql.md

  [comment-analyzer] sql.md vs all other files

  sql.md uses upload_transform_task_artifact and send_transform_message, while every other file (SKILL.md, dotnet.md, tools.md) uses upload_artifact and send_message. Neither alternative name appears in
  tools.md (the canonical tool reference).

  Fix: Unify tool names throughout sql.md to match the canonical reference.

  4. IAM policy attachment silently skips unrecognized identity types

  [silent-failure-hunter] custom-cli-reference.md, custom-remote-execution.md

  The if/elif pattern for attaching IAM policies has no else clause. If the caller ARN doesn't match :user/ or :role/ patterns (e.g., federated user, root), nothing is attached and execution continues as if it
   succeeded. The user discovers this only when Lambda calls fail with AccessDeniedException much later.

  Fix: Add an else clause that warns about the unrecognized identity type.

linchenk369:
Thanks for the review! Good catches.

**Issue 1 — `INTERFACE_METADATA` in `generate_codex_manifests.py`:**
We'll add the `aws-transform` entry.

**Issue 2 — Prod URL in `sql.md` line 157:**
Removed the hardcoded prod URL. Replaced with an example placeholder (`https://xxxxxxxx.transform.us-east-1.on.aws`) and updated the instruction to ask the user for their URL.

**Issue 3 — Inconsistent tool names in `sql.md`:**
Fixed. `upload_transform_task_artifact` → `upload_artifact` and `send_transform_message` → `send_message`, aligned with the MCP server's canonical tool names.

**Issue 4 — IAM policy `else` clause:**
This is owned by the custom workload team. We'll flag it to them.

Will push a fix to the PR to address 1/2/3.

linchenk369 (approved):

youtuyy:
@krokoko @linchenk369 All issues addressed in latest push https://github.com/awslabs/agent-plugins/compare/419afbd4aa7d80b4248ada5d2b7d578419977568..c25bf74ccf93b72407370b02643017624984a727

krokoko (approved):

PR #187, opened 2026-06-09
feat(sagemaker-ai): update skills and bump version to 1.2.0

bhavana-nair:
SageMakerAIAgentSkills (mainline d613419):

  - Added model-selection and sdk-getting-started skills, and finetuning-technique skills. Removed finetuning-setup (replaced by model-selection + finetuning-technique).
  - Bug fixes across multiple skills. 
  - Add 7 new models (Nemotron 3 Nano 30B, Llama 3.1 8B, 5 VLMs) to benchmarks, licenses, and deployment references

  HyperPodAICapabilities: unchanged

  X-AI-Prompt: Sync latest skills content and bump version for release

  X-AI-Tool: Kiro CLI

#### Changes

Sync latest skills content and bump version for release

#### CI note: 
Semgrep may flag `OpenAI` references in benchmark tables (`detect-generic-ai-oai`). This is a false positive — the rule is already excluded in mise.toml but the GH Actions workflow does not mirror
those exclusions.
#### Acknowledgment


aws-pavankks (dismissed):

joshuatowner (dismissed):

krokoko:
## Automated review summary

Thanks for the thorough sync, @bhavana-nair! 🙏 I ran the repo's gates plus a few specialized review passes over the diff. Everything below was verified directly against the changed files, and I confirmed the Critical items are lines **newly added** by this PR (not pre-existing).

### Automated gates — all passing ✅

| Gate | Result |
|---|---|
| `lint:manifests` (ajv) | ✅ all valid; versions consistent at `1.2.0` |
| `lint:cross-refs` | ✅ 0 errors |
| `lint:md` / `fmt:check` | ✅ clean |
| security (bandit/semgrep/gitleaks/checkov/grype) | ✅ no leaks, 0 vulns |

The Semgrep `OpenAI` false-positive you noted in the description checks out — all flagged Python lines are already `nosemgrep`-annotated.

### 🔴 Critical — `NameError` crashes in copy-paste templates

These all live in the newly-added "save manifest" cells, so they fail on **every run, at the final step, after the expensive deploy/eval has already succeeded** — which makes them maximally confusing for a user to debug.

1. **`model-deployment/code_templates/deploy-oss-bedrock.py:119`** — `"model_id": model_id`, but `model_id` is never assigned (the ARN is in `model_arn`, line 79). → `NameError`. Suggest using `model_arn`.
   - Related: on a `Failed`/`Stopped` import the loop `break`s without setting `model_arn`, then the inference cell calls `invoke_model(modelId=model_arn, …)` → a second `NameError` that masks the real failure message. Consider `raise RuntimeError(failureMessage)` on that branch.
2. **`model-deployment/code_templates/deploy-nova-bedrock.py:60`** — uses `json.dumps(...)` but never `import json` (the sibling `deploy-nova-sagemaker.py` does). → `NameError`. Add `import json`.
3. **`model-evaluation/code_templates/custom_scorer_evaluator.py:79,82`** — uses `Path(...)` and `json.dumps(...)` but imports neither (the sibling `llmaaj_evaluator.py` does). → `NameError`. Add `import json` and `from pathlib import Path`.

### 🟡 Important

4. **`finetuning/code_templates/rlaif_builtin.py:7`** — pins `'sagemaker>=3.7.1'` while all four sibling templates pin `'sagemaker>=3.7.1,<4.0'`. As written it would silently install an untested SDK 4.x. Suggest adding `,<4.0` for consistency.
5. **`model-evaluation/scripts/nova_reward_function_source_template.py:343-351`** — the `except Exception` assigns `aggregate_reward_score: 0.0` with no logging. A bug in a user's reward function then silently corrupts the RL training signal for every affected sample (no crash, just bad convergence and wasted GPU hours). Suggest at minimum `print(...)`-ing the exception so it surfaces in CloudWatch, ideally re-raising.
6. **No eval coverage for the routing changes.** The core behavioral change here is skill routing (per `planning/references/skill-routing-constraints.md`), driven entirely by the three new skills' `description` frontmatter — but `sagemaker-ai` has no eval suite, whereas sibling `databases-on-aws` has trigger + functional evals. Not a merge blocker, but a follow-up `tools/evals/sagemaker-ai/` (mirroring the databases-on-aws structure, trigger evals first) would catch routing regressions.

### 🟢 Suggestions (non-blocking)

- `reward_function_source_template.py:241` catches broadly and returns an opaque `400` with no CloudWatch logging.
- `get_recipes.py` / `get_model_names.py`: no error handling around the boto3 calls (raw tracebacks vs. actionable messages); `format_detector.py` is a good in-repo pattern to follow. The module-scope `boto3.client()` also makes the routing-critical parsing logic hard to unit-test.
- `sdk-getting-started/SKILL.md` is the only one of the 11 skills missing the `metadata.version` block (cosmetic — the schema doesn't require it).
- The two reward templates use deliberately opposite `body` contracts (`json.dumps([result])` vs. a raw list) — plausibly correct for different frameworks, but a one-line confirming comment would prevent a copy-paste foot-gun.

### Strengths 👏

- Version bump is consistent across all three manifests — no stray `1.1.0`.
- All renames (`scripts/` → `code_templates/`, `finetuning-setup` → `model-selection`/`finetuning-technique`) resolve cleanly; SKILL.md and reference paths all point to existing files.
- `format_detector.py`, `llmaaj_evaluator.py`, and the `deploy-*-sagemaker.py` templates are clean — pagination/dedup/format-classification logic all looks sound. Scope is tight, with no accidental edits to other plugins.

### Suggested path to merge

1. Fix the four Critical `NameError`s (#1–3) — one line each, but they break every user.
2. Address #4 (version pin) and #5 (silent `0.0` reward) before merge.
3. File a follow-up for the eval suite (#6).

aws-pavankks (approved):

krokoko (approved):

PR #168, opened 2026-05-16
feat(dsql): Add PostgreSQL schema conversion and migration references

pyraenix:
Extends the DSQL skill with PostgreSQL-to-DSQL migration knowledge that complements dsql_lint.

## What's added
- 9 reference files in `references/pg-migrations/` (type mapping, PL/pgSQL patterns, FK replacement, index conversion, schema objects, function compatibility, OCC retry, data migration, multi-region)
- 3 ORM guides in `references/orm-guides/` (Django, Hibernate, Rails)
- 13 new evals in `pg_migration_evals.json` (70/70 expectations pass at 100%)
- Updated SKILL.md with new workflows (9: Full PG→DSQL Migration, 10: ORM Migration)

## Coverage
All 16 items from the gap analysis are implemented and tested:
ENUM→CHECK, PL/pgSQL→SQL, triggers, GIN/GiST/BRIN→btree, partial indexes, expression indexes, materialized views, COLLATE C, multi-schema, FK→validation functions, roles/IAM, OCC retry, ORM adapters, COPY→INSERT, uuid_generate_v4→gen_random_uuid, lastval→currval.

## Design principle
No duplication with dsql_lint. The linter handles mechanical fixes. The skill handles semantic conversions the linter cannot automate (code generation, architectural guidance, ORM patterns).

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/agent-plugins/blob/main/LICENSE).

PR #188, opened 2026-06-09
fix(dsql): address PR #168 post-merge review

amaksimo:
## Summary

Addresses the post-merge review on #168 ([comment](https://github.com/awslabs/agent-plugins/pull/168#issuecomment-4662771095)). Each item was verified against the current files on `main`, awslabs source repos, and (for the COLLATE lint behavior) the live `dsql_lint` MCP tool before applying the fix.

### Correctness (high-confidence findings)
- **Hibernate Maven artifact**: `aurora-dsql-hibernate` → `aurora-dsql-hibernate-dialect` ([upstream README](https://github.com/awslabs/aurora-dsql-orms/blob/main/java/hibernate/README.md))
- **Node connector packages**: split into `@aws/aurora-dsql-node-postgres-connector` (`pg`) and `@aws/aurora-dsql-postgresjs-connector` (`Postgres.js`) ([upstream package.json](https://github.com/awslabs/aurora-dsql-connectors/blob/main/node/node-postgres/package.json))
- **`@Retryable` guidance**: `PessimisticLockException` is wrong for OCC. Recommend the [aurora-dsql-jdbc-connector](https://github.com/awslabs/aurora-dsql-connectors/tree/main/java/jdbc) or a manual handler matching on SQLSTATE 40001
- **SKILL.md description/tags**: drop "PL/pgSQL transpilation", `plpgsql`, `trigger` (DSQL rejects `LANGUAGE plpgsql` and `CREATE TRIGGER`)
- **MCP tool naming**: standardize on `dsql_lint` (underscore); the hyphen form is the CLI binary, not the MCP tool

### Skill structure
- **`SKILL.md` Reference Files**: drop duplicated per-file `###` subsections; surface data-loading and query-plan refs in proper tables
- **Workflow 2**: restore the "MUST issue each `ALTER TABLE` in its own `transact` call" bullet (DSQL rejects multi-DDL transactions)

### Silent-failure fixes
- **`pg-migrations/index-conversion.md`**: JSONB GIN backfill switched from `ADD COLUMN` + `UPDATE` to `GENERATED ALWAYS AS (...) STORED` — closes the race window where rows inserted between the two statements would silently be left out of the index
- **`pg-migrations/fk-replacement.md`**: add an explicit caller pattern showing the `validate_fk_*()` result MUST gate the INSERT in the same transaction (DSQL only supports `LANGUAGE sql`, so enforcement happens in the application layer)
- **`pg-migrations/schema-objects.md`**: warn that removing `SECURITY DEFINER` without auditing GRANTs can convert an enforced boundary into a silent permission bypass

### Consistency
- **Eval IDs**: `pg_migration_hallucination_evals.json` moved to 4xx range to remove the 300/301/302 collision with `pg_migration_evals.json`; results table updated
- **`multi-region.md`**: reword "Regions per cluster: 2 + 1 witness" to "two peered clusters in two endpoint Regions plus one shared witness Region (no endpoint)"
- **COLLATE wording**: `development-guide.md`, `type-mapping.md`, `schema-objects.md` unified on "`dsql_lint(fix=true)` auto-strips COLLATE" — verified via the MCP tool (rule `collation`, status `fixed`)

### Items not actioned (and why)
- **Anchor slugs (item 17)**: re-checked the ToCs in `schema-objects.md` and `index-conversion.md`; the slugs correctly match what GFM produces from the heading text
- **PR description / commit history (items 12, 14)**: post-merge, not fixable retroactively
- **Workflow 9 duplicate (item 1)**: already fixed by #186

## Test plan

- [ ] Reviewer eyeballs the SKILL.md `Reference Files` section for any remaining duplicated entries
- [ ] Reviewer confirms the new node connector table reads correctly across the two driver columns
- [ ] Run `dsql_lint(fix=true)` against a `COLLATE "C"` DDL to confirm the auto-fix wording matches actual lint behavior

---


plugins/databases-on-aws/skills/dsql/references/orm-guides/overview.md
```diff
@@ -5,12 +5,12 @@ names and configuration not available in general documentation.
```
krokoko, line 5: Don't forget to bump the plugin version :) I'll add a CI check for it to make it easier (and maybe a script to do it on your behalf so you don't have to think about it)

amaksimo: Thanks! I always forget this lol

plugins/databases-on-aws/skills/dsql/SKILL.md
```diff
@@ -1,9 +1,9 @@
...
+description: "Build with Aurora DSQL — manage schemas, execute queries, handle migrations, diagnose query plans, load data, and develop applications with a serverless, distributed SQL database. Covers IAM auth, multi-tenant patterns, MySQL-to-DSQL and PostgreSQL-to-DSQL schema conversion, FK replacement code generation, OCC retry patterns, ORM migration (Django/Hibernate/Rails), DDL operations, query plan explainability, SQL compatibility validation, and bulk data loading. Triggers on phrases like: DSQL, Aurora DSQL, distributed SQL database, serverless PostgreSQL-compatible database, migrate to DSQL, DSQL query plan, DSQL EXPLAIN ANALYZE, DSQL ENUM, DSQL foreign key, DSQL OCC retry, DSQL multi-region, DSQL JSONB, DSQL GIN index, aurora-dsql-loader."
```
anwesham-lab: we should probably prefer "load into DSQL" over the loader assuming people don't know about the loader, but that's follow up

plugins/databases-on-aws/skills/dsql/SKILL.md
```diff
@@ -69,76 +69,20 @@ Load these files as needed for detailed guidance:
...
+| Reference                                     | When to Load                                             | Contains                                                                                  |
+| --------------------------------------------- | -------------------------------------------------------- | ----------------------------------------------------------------------------------------- |
+| [data-loading.md](references/data-loading.md) | Planning or running bulk loads with `aurora-dsql-loader` | Fresh-vs-warm partitions, resume/retry, `--on-conflict` semantics, throughput diagnostics |
 
-#### [ddl-migrations/batched-migration.md](references/ddl-migrations/batched-migration.md)
-
-**When:** Load when migrating tables exceeding 3,000 rows
-**Contains:** OFFSET-based and cursor-based batching patterns, progress tracking, error handling
-
-### MySQL Migrations (modular):
```
anwesham-lab, line 114: what was the purpose of cutting these? also we should definitely still run the pre-existing evals again to validate that there isn't a regression when we make key alterations to the top level skill.

amaksimo: Re-running and verifying — net result: no regression, no content lost.

**Duplication proof (the cuts).** Of the 18 `###`/`####` subsections removed: 16 were verbatim duplicates of rows already in the pre-existing Core / MCP / DDL Migrations / MySQL Migrations category tables; 2 (`data-loading.md`, `query-plan/*`) were orphans not in any table — PR #188 promoted them into new "Data Loading" and "Query Plan Explainability" tables, so they survive; 1 was a wrong stub (a "Query Plan Explainability" heading describing troubleshooting content). Net file-link delta: zero.

**Eval re-run** (`fix/dsql-pr168-followup` HEAD):

| Suite | Result | vs pre-#188 |
| --- | --- | --- |
| Tier 2 functional (`evals.json`) | 39 / 42 (93%) | No regression |
| PG migration (`pg_migration_evals.json`) | 71 / 76 (93%) | No regression |
| Hallucination (`pg_migration_hallucination_evals.json`) | 14 / 14 (100%) | Same as #168 baseline |
| Tier 1 trigger (`trigger_evals.json`) | 15 / 31 (48%) | **Identical to pre-#188** |

The trigger score looks low but it's the noise floor — re-ran the same suite against the pre-#188 description (commit `6477dfd~1`); also 15/31, with the **exact same 16 failing queries** (`comm -23` returned empty). I tried an aggressive trigger-phrase expansion during the investigation; it didn't move the score, so I reverted it.

Happy to share the raw per-eval JSONs if useful — keeping them out of the repo since they'd be a per-PR snapshot rather than a maintained eval reference.

amaksimo: The tier 1 trigger, I can look at improving the results in a follow-up pr.

plugins/databases-on-aws/skills/dsql/SKILL.md
amaksimo: just did it here.

anwesham-lab (approved):

krokoko (approved):

PR #171, opened 2026-05-19
feat: Add AWS Elastic Beanstalk to deploy-on-aws plugin

azpaulp:
Add a new elastic-beanstalk skill to the deploy-on-aws plugin covering web and worker applications. EB is positioned as an application management service (AWS manages lifecycle) vs ECS/EKS as infrastructure management services (user manages lifecycle).

#### Changes

New files:
- skills/elastic-beanstalk/SKILL.md — routing, workflow, defaults
- skills/elastic-beanstalk/references/platforms.md — platform detection
- skills/elastic-beanstalk/references/configuration.md — ebextensions, hooks

Modified existing files:
- defaults.md — override triggers and Heroku migration row
- security.md — VPC, SG, IAM, logging entries for EB
- cost-estimation.md — EB pricing (free service, underlying costs)
- deploy SKILL.md — principle and reference link

Testing:
- `markdownlint-cli2`: 0 errors in changed files
- `dprint check`: passes clean (no self-mutations)
- `validate-references.py`: 0 broken links
- `validate-size.py`: SKILL.md = 160 lines / 1954 tokens (rated "ideal")
- `validate-cross-refs.cjs`: 0 errors
- `End-to-end test`: Deployed a .NET 10 ASP.NET Core chatbot (Bedrock integration) to Elastic Beanstalk using only AWS CLI commands from the skill. Verified health endpoint responded 200 OK and Bedrock invocation succeeded via IAM instance profile. Two issues found and fixed in the skill: pre-built artifact requirement for .NET, and reverse proxy port configuration.

#### Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/agent-plugins/blob/main/LICENSE).

PR #190, opened 2026-06-10
fix(deploy-on-aws): clarify serverless vs managed-server distinction …

azpaulp:
…in EB skill

Small tweaks on top of https://github.com/awslabs/agent-plugins/pull/171

"Don't want to manage servers" should route to EB, not Lambda. Serverless imposes a different programming model (stateless functions, cold starts, event-driven) — it's not just "managed servers." This change:
- Fixes trigger text: "containers" → "container orchestration" (EB supports Docker)
- Expands the NOT-right-choice bullet to explain why serverless is different
- Adds a paragraph to Key Distinction explaining the EB-vs-Lambda boundary

#### Acknowledgment


krokoko (approved):

scottschreckengaust (approved):
LGTM

PR #193, opened 2026-06-11
Fix command to install skills for Kiro agent

Krxtopher:
This pull request updates the installation instructions in the `plugins/sagemaker-ai/README.md` to clarify how to add skills for other agents using the Skills CLI.

#### Related

N/A

#### Changes

Documentation update:

* Changed the Skills CLI command to use `--skill *` instead of `--all` when installing all skills for an agent, providing clearer and more accurate guidance for users.The previous command installed all skills for *all* coding agents, resulting in many irrelevant dot directories being created and populated with skill copies (.claude, .aider-desk, .kode, etc.)

#### Acknowledgment


Krxtopher:
There may still be an error in the install command instructions. I'm continuing to investigate.

PR #209, opened 2026-06-30
feat: sync AWS Transform agent-plugin from ATXKiroPower (post cli-3.3.0)

jdonboch:
Promotes `ATXKiroPower/mainline` changes merged since the `cli-3.3.0` tag.

**Themes:** Continuous Modernization EC2/Batch execution rewrite (CFN + SSM, multi-worker, memory caps, prechecks), scheduling, security agent skill (admin/executor split), mainframe-reimagine workflow, remediation + custom-TD, source handling (GitHub/GitLab/Bitbucket), reporting, telemetry, and markdown lint/format cleanup.

**Version bump:** 1.1.2 → 1.2.0 (minor; feature release) across `.claude-plugin/plugin.json`, `.codex-plugin/plugin.json`, and the `aws-transform` entry in `.claude-plugin/marketplace.json`.


plugins/aws-transform/.claude-plugin/plugin.json
```diff
@@ -1,35 +1,8 @@
 {
-  "author": {
-    "name": "Amazon Web Services"
-  },
-  "description": "Migrate, modernize, and upgrade codebases to AWS. Transforms .NET Framework to .NET 8/10, mainframe COBOL to Java, VMware VMs to EC2, SQL Server to Aurora, and upgrades Java/Python/Node.js versions and AWS SDKs. AWS Transform - continuous modernization analyzes codebases for tech debt, security issues, and upgrade opportunities, then remediates them.",
-  "homepage": "https://github.com/awslabs/agent-plugins",
-  "keywords": [
```
dhasani23, line 7: I don't think we should delete all of these keywords

plugins/aws-transform/skills/aws-transform/references/mainframe-reimagine-ddd.md
```diff
@@ -0,0 +1,316 @@
+# Phase 2: DDD Bounded Context Analysis
```
carolabadeer: Are these mainframe-* file changes owned by us?

plugins/aws-transform/.claude-plugin/plugin.json
jdonboch: agree, will fix, working with agent to understand why this happened

plugins/aws-transform/skills/aws-transform/references/mainframe-reimagine-ddd.md
jdonboch: they are not but @lqashe has given us the go away to release these changes with ours

plugins/aws-transform/skills/aws-transform/scripts/verify-traceability.py
```diff
@@ -0,0 +1,705 @@
...
+def generate_html(result: dict) -> str:
+    """Produce a self-contained HTML dashboard string."""
+    s = result["summary"]
+    gen_time = result["generated_at"]
+
+    rules_pct = _pct(s["implemented_rules"], s["total_rules"])
+    reqs_pct = _pct(s["implemented_requirements"], s["total_requirements"])
+    overall_total = s["total_rules"] + s["total_requirements"]
+    overall_impl = s["implemented_rules"] + s["implemented_requirements"]
+    overall_pct = _pct(overall_impl, overall_total)
```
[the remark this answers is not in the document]
jdonboch: minor, will fix in future release

carolabadeer (dismissed):

ntarakad-aws (dismissed):

carolabadeer (approved):

scottschreckengaust (approved):
## Approve — with nits

Reviewed the substantive part of this sync: the embedded CloudFormation/bash in `continuous-modernization-ec2-execution.md` (default `WorkerCount` 5→1, the new memory precheck, per-container `--memory` caps, `--restart unless-stopped` → `on-failure:3`, the rewritten health-check loop, the laptop-side/`ssm_run` memory math, and the new Secrets Manager credential-validation step), plus the skill-instruction reflows in `source`/`remediation`/`schedule`/`batch-execution`.

The new logic is carefully built. I verified the high-risk mechanics rather than assuming, and they hold up. The items below are hardening nits, not correctness regressions — hence approving.

### ✅ Verified correct
- **Memory math** `(MEM_TOTAL_MB - 4096) / WorkerCount` with a 2048 floor: the precheck (`WorkerCount*2048 + 4096`) guarantees the floor never over-commits the host; `--memory-swap == --memory` correctly disables overflow swap.
- **`!Sub` escaping**: every `${!bash_var}` vs `${CfnRef}` is correct, including the deliberate `wait "${!PIDS[@]}"` (renders to literal `${PIDS[@]}` for bash).
- **Health-check terminal logic** `dead OR (exited AND (ExitCode==0 OR RestartCount>=3))` exactly matches its comment and correctly tolerates the transient `exited` snapshot between `on-failure:3` retries.
- **`WorkerCount=1` default is consistent everywhere** — CFN default, MinValue/MaxValue, provisioning script auto-sizing, decision tables, and prose. No stale references to the old default of 5.

### ⚠️ Nits — silent-failure hardening in the new code (non-blocking)

1. **Precheck can fail _open_ if `MemTotal` reads empty.** `MEM_TOTAL_MB=$(( $(awk '/MemTotal/{print $2}' /proc/meminfo) / 1024 ))` — if `awk` yields nothing, the arithmetic errors but `set -e` does **not** abort; `[ "" -lt REQUIRED_MB ]` is then false, so UserData prints "PASSED" and continues, and the empty value later poisons `--memory="m"`. This defeats the precheck's stated purpose (fail in seconds). Suggest validating `MEM_TOTAL_MB` is a positive integer and `exit 1` with a clear message otherwise.

2. **`--restart on-failure:3` leaves a container permanently down post-provision with no watchdog.** Correct at provision time (a clean exit surfaces via the CreationPolicy timeout). But after `CREATE_COMPLETE`, a runtime OOM-loop that exhausts 3 retries — or a clean exit — leaves the container dead silently; the next job's `docker exec` just fails. Consider having the Step 9 status path verify `State.Status=running` and message clearly ("container down — retries exhausted") rather than reporting a generic "no AID / still pending".

3. **Laptop-side `ssm_run` memory-math errors are swallowed.** An empty remote `MEM_TOTAL_MB`, or `WORKER_COUNT=0`, makes the remote `docker run` fail — but the error lands only in `StandardErrorContent`, which the `ssm_run` helper never reads (it fetches `StandardOutputContent` only), so the flow proceeds as if the container launched. Note the `${WORKER_COUNT:-1}` fallback guards unset/empty but **not** a literal `0`. Suggest validating remotely and surfacing `StandardErrorContent` (and the command `Status`) on non-`Success`.

4. **New Step 7 credential snippet masks create failures.** `create-secret … 2>/dev/null || put-secret-value …` hides an `AccessDenied`/KMS failure and falls through to `put-secret-value`, which then fails with a misleading `ResourceNotFoundException`. The sibling `describe-secret … 2>&1` in the same step surfaces errors correctly — worth making these consistent (only fall back on the specific already-exists condition).

### 📝 Maintainability advisory
No comment contradicts its code. The one real drift risk is **magic numbers duplicated in prose**: `2048`/`4096` (three spots), `on-failure:3` coupled to the health check's `-ge 3`, and the computed **"300s"** (actually `seq 1 60 × sleep 5`) stated as a literal. All correct today, but nothing keeps them in sync — consider hoisting shared constants into named vars or adding "keep in sync with X" pointers. Also worth confirming the troubleshooting table's **port 8081** is accurate (it's unverifiable from the file alone).

### 🔭 Out of scope (not introduced by this PR)
Two items I initially surfaced trace to **pre-existing code this PR doesn't touch** (the remediation fan-out's `--ids`/`--repo` interaction, and the `ssm_submit` fire-and-forget launch not verifying its sentinel). Flagging for a possible separate follow-up — not blockers here.

Nice, defensive work on the memory-capping and health-check rewrite. 👍

<sub>Review assisted by Claude Code (pr-review-toolkit).</sub>

Issue #221, opened 2026-07-07
Add zizmor to CI/security scans for GitHub Actions static analysis


scottschreckengaust:
## Summary

Wire [zizmor](https://docs.zizmor.sh/) — a purpose-built static analyzer for GitHub Actions — into the repo's tooling so workflows are scanned for security issues (template/script injection, dangerous triggers, credential persistence, excessive permissions, unpinned actions, etc.).

## Motivation

While remediating a script-injection vuln in `merge-prevention.yml` (CWE-78, PR #222) we found zizmor is **not installed or referenced anywhere** — not in `mise.toml`, `.pre-commit-config.yaml`, or CI. It would have flagged that finding automatically, and it caught a latent invalid-`if:`-expression bug during that work.

## Scope — all three layers (match the existing scanner pattern)

Every other scanner (bandit, semgrep, checkov, gitleaks) lives in all three layers; zizmor should too:

- [ ] **CI** — add a `zizmor` job to `.github/workflows/security-scanners.yml`, mirroring the bandit/gitleaks pattern (upload results to code scanning).
- [ ] **mise** — add `security:zizmor` task; include in the `security` aggregate (which `build` runs).
- [ ] **pre-commit** — add the official hook (`zizmorcore/zizmor-pre-commit`), naturally scoped to workflow files.

## ⚠️ Key obstacle: pre-existing findings

A baseline run (zizmor 1.26.1, default persona, `--offline`) over `.github/workflows/` reports **72 findings: 1 high, 12 medium, 3 low, 7 informational** (rest suppressed). The single **high** is `dangerous-triggers` — `pull_request_target` in `pull-request-lint.yml`.

This means zizmor **cannot be a blocking gate on day one** without first triaging/fixing that backlog. Introduction strategy (pick one):

1. **Advisory first** — run + report (upload SARIF to code scanning) but non-blocking; burn down findings in follow-ups; flip to blocking later. *(Recommended — standard "new scanner vs. existing backlog" playbook.)*
2. **Blocking immediately** — triage all 72 (fix or document per-finding ignores via `.github/zizmor.yml`) in the same PR, then gate. Larger, riskier (touches many unrelated workflows).

## Notes

- CLI emits `plain`/`json` only (no native SARIF in 1.26.1); the official `zizmorcore/zizmor-action` handles SARIF conversion for code-scanning upload.
- Runs with zero repo dependency via `uvx zizmor` (same ephemeral pattern as pytest).
[...]

PR #222, opened 2026-07-07
fix(ci): prevent script injection in merge-prevention workflow

scottschreckengaust:
## What

Remediates **script injection (CWE-78)** in `.github/workflows/merge-prevention.yml`. Untrusted GitHub context values were interpolated directly into `run:` shell scripts via `${{ }}`, where Actions substitutes them into the script text *before* the shell executes — so a crafted branch ref or PR label could inject arbitrary commands into the runner.

## Injection points fixed

| Location | Untrusted value | Risk |
| --- | --- | --- |
| `fail-by-label` echo | `pr_labels` output (derived from user-set PR **labels**) | High — most user-controllable |
| `get-pr-info` | `github.ref`, `github.repository`, `github.event.pull_request.number`, `github.event_name` | Medium |
| step `name:` fields | `${{ env.DO_NOT_MERGE_LABEL }}` | Low (hygiene) |

## How

- Use runner **built-in env vars** (`$GITHUB_EVENT_NAME`, `$GITHUB_REF`, `$GITHUB_REPOSITORY`) instead of interpolating `github` context.
- **Bind** PR number and labels to `env:` and reference them as quoted shell variables (`"$VAR"`) so values are treated as data, not code.
- Quote the `gh api` URL argument.
- Use the already-exported `$DO_NOT_MERGE_LABEL` in `run:` scripts and drop the interpolation from step names.
- Wrap the negated condition as `if: ${{ !contains(...) }}` — the original `if: ! contains(...)` was **invalid GitHub Actions expression syntax** (caught by zizmor's schema validation).

## Verification

- **semgrep**: original file triggers `yaml.github-actions.security.run-shell-injection`; fixed file scans clean.
- **zizmor** (v1.26.1, `--offline`): `No findings to report` on the fixed file — including its `template-injection` audit. The original file failed zizmor's schema validation due to the invalid `if:` syntax noted above.
- YAML parses successfully.

## Follow-up

- zizmor is not yet part of CI — tracked in #221.
- Pre-existing gitleaks findings on an unmerged branch are unrelated to this change and handled separately.

---

### DIFF
--- a/.github/workflows/merge-prevention.yml
+++ b/.github/workflows/merge-prevention.yml
@@ -45,18 +45,19 @@ jobs:
     env:
       GH_TOKEN: ${{ github.token }}
       PR_LABELS_JSON: ${{ toJson(github.event.pull_request.labels.*.name) }}
+      PR_NUMBER_EVENT: ${{ github.event.pull_request.number }}
     steps:
       - name: Get PR info
         id: get-pr
         run: |
-          if [ "${{ github.event_name }}" == "merge_group" ]; then
-            PR_NUMBER=$(echo "${{ github.ref }}" | grep -oP '(?<=/pr-)\d+' || echo "")
-            PR_LABELS=$(gh api repos/${{ github.repository }}/pulls/$PR_NUMBER | jq -c '[.labels[].name] // []')
+          if [ "$GITHUB_EVENT_NAME" == "merge_group" ]; then
+            PR_NUMBER=$(echo "$GITHUB_REF" | grep -oP '(?<=/pr-)\d+' || echo "")
+            PR_LABELS=$(gh api "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER" | jq -c '[.labels[].name] // []')
             echo "::group::Getting Information"
-            gh api repos/${{ github.repository }}/pulls/$PR_NUMBER
+            gh api "repos/$GITHUB_REPOSITORY/pulls/$PR_NUMBER"
             echo "::endgroup::"
-          elif [ "${{ github.event_name }}" == "pull_request" ]; then
-            PR_NUMBER="${{ github.event.pull_request.number }}"
+          elif [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
+            PR_NUMBER="$PR_NUMBER_EVENT"
             PR_LABELS=$(echo "$PR_LABELS_JSON" | jq -c '.')
           fi
           echo "::group::Debug Output Values"
@@ -70,8 +71,9 @@ jobs:
     needs: get-pr-info
     if: always()
     steps:
-      - run: |
-          PR_NUMBER="${{ needs.get-pr-info.outputs.pr_number }}"
+      - env:
+          PR_NUMBER: ${{ needs.get-pr-info.outputs.pr_number }}
+        run: |
           # Default to 0 (allow all) if not set
           if [ -z "$HALT_MERGES" ]; then
             HALT_MERGES=0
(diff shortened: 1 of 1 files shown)


theagenticguy (approved):
## ✅ Approved

Clean, well-scoped remediation of the CWE-78 script-injection in `.github/workflows/merge-prevention.yml`. Diffed both revisions line-by-line, confirmed functional equivalence, and swept the other workflows in `.github/workflows/` for the same vulnerability class.

### Why the remediation is sound
- **Untrusted context moved out of script text.** `github.event_name`/`github.ref`/`github.repository` → runner built-ins `$GITHUB_EVENT_NAME` / `$GITHUB_REF` / `$GITHUB_REPOSITORY`; PR number and labels bound to `env:` and read as quoted `"$VAR"`. Values now reach the shell as data, so a crafted branch ref or an attacker-set label like `$(...)` or `; rm -rf` stays inert. The `fail-by-label` echo of `pr_labels` was the sharpest exposure since PR labels are the most user-controllable input.
- **Behavior preserved.** `PR_NUMBER_EVENT` is a new job-level env binding of `github.event.pull_request.number`, referenced only in the `pull_request` branch — same value the old inline interpolation produced. Control flow, job outputs (`pr-number`/`pr-labels`), and exit codes are unchanged.
- **Correctness bug fixed in passing.** `if: ! contains(...)` was invalid Actions expression syntax; `if: ${{ !contains(...) }}` is valid — that negative branch was likely misbehaving before this change.
- **`gh api` URL is quoted**, closing word-splitting/glob exposure on the interpolated path.

### Verification
- The PR's semgrep result (`yaml.github-actions.security.run-shell-injection` dirty → clean) and zizmor v1.26.1 `--offline` "no findings" are consistent with the diff; the invalid `if:` explains why the original failed zizmor's schema validation.
- Independent sweep of the sibling workflows found no surviving untrusted-context interpolation inside `run:` blocks — the only remaining `${{ }}` uses in scripts are `env.SONAR_TOKEN`, a secret correctly bound to env.

### Non-blocking follow-ups (out of scope for this PR)
- Pin `zizmor` into CI so this class is caught automatically (already tracked in #221).
- Consider `persist-credentials: false` hygiene across the workflow set.

Merge it.

isadeks (approved):
Correct, complete, and minimal security fix following the canonical GitHub hardening pattern (bind untrusted context to `env:`, reference as quoted `"$VAR"`). Verified each change:

- **`get-pr-info`**: `$GITHUB_EVENT_NAME` / `$GITHUB_REF` / `$GITHUB_REPOSITORY` are correct runner built-ins; `PR_NUMBER_EVENT` bound to env and quoted; `gh api` URL quoted. ✓
- **`fail-by-label`**: the debug `echo` now reads `$PR_LABELS` from env instead of interpolating label JSON — this was the genuinely data-plane value flowing into a shell, correctly fixed. ✓
- **`if: ! contains(...)` → `if: ${{ !contains(...) }}`**: right call — a leading `!` is a reserved YAML tag indicator, so the original was invalid; wrapping makes it a valid expression negation. ✓
- Scanned the whole file: no remaining `${{ }}` interpolation of untrusted context inside any `run:` block. The `if:` conditions reference outputs/env but evaluate in GitHub's expression engine via `contains()` (data, not shell), so they're safe. ✓

**Non-blocking notes**
- Label ranking as "High — most user-controllable": in the fork-PR threat model an external contributor can't add labels (needs triage/write), so the exploitability bar is higher than a `head_ref` injection would be. Still the right thing to fix.
- The step `name:` change is cosmetic (names aren't shell-evaluated); minor cosmetic tradeoff if `vars.DO_NOT_MERGE_LABEL` is overridden the name says "do-not-merge" while the error message prints the real value. Acceptable.

Verification (semgrep flags original / clean on fix; zizmor `--offline` no findings incl. template-injection; YAML parses) is appropriate for a workflow change. Approving.

PR #224, opened 2026-07-07
fix(ci): scope gitleaks scan to PR commits, not all branches

### LINKED ISSUE #223: gitleaks CI fails PRs on unrelated branches' content (scan scope too broad)
## Summary

The `gitleaks` job in `.github/workflows/security-scanners.yml` fails pull requests because of secrets that exist **only on other, unmerged branches** — not in the PR, not in `main`.

## Root cause

The step runs:

```
gitleaks git --config=.gitleaks.toml --baseline-path=.gitleaks-baseline.json ... .
```

with **no `--log-opts` range**. In git mode, `gitleaks` walks the entire object graph reachable from **all refs** in the clone. Because `actions/checkout` uses `fetch-depth: 0`, the runner fetches every branch (`+refs/heads/*:refs/remotes/origin/*`), so the scan reaches sideways into unrelated branches.

## Evidence

Scoped local scans (gitleaks 8.30.0, same config + baseline):

| Scope | Result |
| --- | --- |
| `origin/main` full history | ✅ no leaks (117 commits) |
| any current PR branch history | ✅ no leaks |
| all refs (gitleaks default) | ❌ **3 leaks** |

The 3 findings (`slack-webhook-url` ×2, `private-key` ×1) are all in `tools/semgrep/r-all.active.yaml` — a file that **does not exist in `main`'s tree**. It lives only on `feat/vendor-semgrep-rules` (open, unmerged PR #220), where they are example patterns inside the vendored Semgrep `r/all` ruleset (false positives).

CI confirms it: the gitleaks job log reports `169 commits scanned` (far more than any single PR branch) → `leaks found: 3`. Every PR opened today fails gitleaks identically, regardless of its content.

## Proposed fix

Scope the scan to the PR's own commits by adding `--log-opts`:

```yaml
- name: Run gitleaks (PR range)
  id: gitleaks
  run: |
    set +e
    BASE="origin/${{ github.base_ref || 'main' }}"
    git fetch --no-tags origin "${{ github.base_ref || 'main' }}"
    gitleaks git --config=.gitleaks.toml \
      --baseline-path=.gitleaks-baseline.json \
      --log-opts="${BASE}..HEAD" \
      --report-path=gitleaks-report_sarif.json --report-format=sarif .
    GITLEAKS_EXIT=$?
    set -e
    echo "exit_code=$GITLEAKS_EXIT" >> "$GITHUB_OUTPUT"
    exit 0
```

`BASE..HEAD` scans only the commits the PR introduces, so a PR can never fail on another branch's content.

### Trade-off / follow-up

PR-range scanning only inspects new commits, so it won't re-catch a secret already in `main`. A separate push/schedule-triggered job should scan `main` in full. (The same broad-scope pattern also exists in the local `mise` `security:gitleaks` task — worth aligning.)

## Notes

- The real slack-webhook/private-key findings on PR #220 should be allowlisted/baselined in that PR (they are example rule fixtures, not live credentials).
- Discovered while remediating a script-injection issue in `merge-prevention.yml` (PR #222); that PR's red gitleaks check is caused by this scan-scope bug, not by its own change.

scottschreckengaust:
Fixes #223

## Problem

The `gitleaks` job in `.github/workflows/security-scanners.yml` runs `gitleaks git ... .` with **no commit range**. In git mode, gitleaks walks the entire object graph reachable from **all refs** in the clone. Because `actions/checkout` uses `fetch-depth: 0`, the runner fetches every branch — so the scan reaches sideways into unrelated branches.

**Consequence:** pull requests fail on secrets that exist only on other, unmerged branches. Concretely, PR #222's gitleaks check is red because of example rule fixtures (`slack-webhook-url`, `private-key`) in `tools/semgrep/r-all.active.yaml` on the unmerged `feat/vendor-semgrep-rules` branch (PR #220) — a file that isn't in `main` or in PR #222 at all. The CI log shows `169 commits scanned` (far more than the PR's own commits) → `leaks found: 3`.

The local `mise run build` (`security:gitleaks` task) had the identical broad-scope problem and failed the same way.

## Fix

**CI** (`security-scanners.yml`) — scope the scan by event type:

| Event | Range | Rationale |
| --- | --- | --- |
| `pull_request` | `origin/<base_ref>..HEAD` | Only the commits the PR introduces — cannot fail on another branch's content |
| `push` / `schedule` / `workflow_dispatch` | `HEAD` | Full history reachable from HEAD — preserves complete coverage of `main` |

**Local** (`mise.toml` `security:gitleaks` task) — add `--log-opts=HEAD` so the local scan matches CI scope (current-branch history only, not sideways branches).

`github.base_ref` is bound to an `env:` var and referenced as `"$PR_BASE_REF"` (never interpolated into the shell) to avoid the script-injection class being addressed in #222.

## Verification

- YAML parses.
- **Local `mise run security:gitleaks` now passes**: 119 commits scanned (HEAD ancestry) vs. 221 (all refs) → `no leaks found`.
- Scoped scans confirm isolation: `origin/main` history → 0 leaks; all-refs default → 3 leaks (the other branch's fixtures).
- `checkout` already fetches full history (`fetch-depth: 0`), so `origin/<base_ref>` is present without an extra fetch.

## Notes / follow-ups

- Draft until reviewed — narrowing scan scope is a deliberate security trade-off. Full-history coverage of `main` is retained via the push/schedule branch of the conditional (CI) and remains available locally on `main`.
- The example secrets themselves should be allowlisted/baselined on PR #220 (they are Semgrep rule fixtures, not live credentials).
- Related: zizmor CI integration (#221), workflow script-injection fix (#222).

---

### DIFF
--- a/.github/workflows/security-scanners.yml
+++ b/.github/workflows/security-scanners.yml
@@ -42,11 +42,24 @@ jobs:
           curl -sSfL "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz" | tar -xz;
           sudo mv gitleaks /usr/local/bin/;
           gitleaks --version;
-      - name: Run gitleaks (full history)
+      - name: Run gitleaks
         id: gitleaks
+        env:
+          # On pull_request, restrict the scan to commits the PR introduces so
+          # it cannot fail on secrets that live only on other, unmerged branches.
+          # On push/schedule/dispatch, scan the full history reachable from HEAD.
+          PR_BASE_REF: ${{ github.base_ref }}
         run: |
           set +e
-          gitleaks git --config=.gitleaks.toml --baseline-path=.gitleaks-baseline.json --report-path=gitleaks-report_sarif.json --report-format=sarif .
+          # checkout uses fetch-depth: 0, so the base branch is already present
+          # as origin/<base_ref>; no extra fetch is needed.
+          if [ "$GITHUB_EVENT_NAME" = "pull_request" ]; then
+            LOG_OPTS="origin/$PR_BASE_REF..HEAD"
+          else
+            LOG_OPTS="HEAD"
+          fi
+          echo "::notice::gitleaks scanning range: $LOG_OPTS"
+          gitleaks git --config=.gitleaks.toml --baseline-path=.gitleaks-baseline.json --log-opts="$LOG_OPTS" --report-path=gitleaks-report_sarif.json --report-format=sarif .
           GITLEAKS_EXIT=$?
           set -e
           echo "exit_code=$GITLEAKS_EXIT" >> "$GITHUB_OUTPUT"
--- a/mise.toml
+++ b/mise.toml
@@ -123,8 +123,11 @@ run = [
 
 [tasks."security:gitleaks"]
 description = "Run GitLeaks"
+# Scan only history reachable from HEAD (the current branch), not every ref in
+# the clone. Without --log-opts, gitleaks walks all refs and fails on secrets
+# that exist only on other, unmerged branches. Matches the CI scan scope.
 run = [
-  "gitleaks git --config=.gitleaks.toml --baseline-path=.gitleaks-baseline.json"
+  "gitleaks git --config=.gitleaks.toml --baseline-path=.gitleaks-baseline.json --log-opts=HEAD",
 ]
 
 [tasks."security:grype"]


theagenticguy (approved):
**Approved.**

Clean, correctly-scoped fix for #223. Verified end to end:

**Range logic is correct on every event path**
- `pull_request`: `origin/$PR_BASE_REF..HEAD` scopes the scan to exactly the commits the PR introduces. With `actions/checkout@v6` + `fetch-depth: 0` the base is already present as `origin/<base_ref>`, so the "no extra fetch needed" comment holds. Sideways/unmerged branches are excluded by construction — this is precisely what was leaking PR #220's fixtures into unrelated PRs.
- `push` / `schedule` / `workflow_dispatch`: `HEAD` walks full ancestry, so `main` keeps complete coverage. The trade-off called out in the description (PR-range won't re-catch a secret already in `main`) is real but correctly mitigated by this branch of the conditional.
- `mise.toml` `--log-opts=HEAD` matches the CI local scope. TOML parses clean.

**Injection-safe**
`github.base_ref` is bound to `PR_BASE_REF` and referenced as `"$PR_BASE_REF"`, never interpolated into the shell — consistent with the hardening in #222.

**CI**
All 24 checks green, including the `gitleaks` job itself now running on the scoped range (`no leaks found`), plus SonarQube/CodeQL/Semgrep.

Non-blocking: none. Agree with the follow-up to baseline the example fixtures on #220. Ship it.

yananliw (approved):
LGTM

isadeks (approved):
Correct root-cause fix, safe implementation, well documented. The diagnosis is right: gitleaks' `git` source defaults to `--all` when `--log-opts` is empty, so with `fetch-depth: 0` it walks every ref — scoping via `--log-opts` is exactly the right lever.

**Strengths**
- Injection-safe: `github.base_ref` is bound to `env:` and only dereferenced as `"$PR_BASE_REF"`, never interpolated into the `run:` block — consistent with the #222 hardening.
- Coverage tradeoff is sound and reversible: full-history coverage of `main` is preserved on push/schedule, and secrets added *within the PR's own commits* still sit in an intermediate commit inside the range, so per-commit scanning still catches them.
- Clear comments + `::notice::` echoing the effective range.

**Non-blocking suggestions**
1. Consider `pull_request.base.sha` for the range (as the sibling `semgrep` job already does): `LOG_OPTS="${BASE_SHA}..HEAD"`. It's the exact fetched commit — immune to `origin/main` advancing and not dependent on the `origin/<base_ref>` remote-tracking ref being materialized. Worth confirming `origin/main` actually resolves in a real PR run, since the runner checks out a detached merge ref and the PR's verification was done locally.
2. The `mise.toml` comment says `--log-opts=HEAD` "Matches the CI scan scope" — it matches only the push/schedule scope, not the PR scope (`origin/<base_ref>..HEAD`). Suggest "Matches the CI push/schedule scan scope."

Both are optional polish. Approving.

PR #227, opened 2026-07-07
feat(build): run unit tests in build via pytest

### LINKED ISSUE #226: Establish a Python unit-test framework wired into `mise run build`
## Gap

The repository has **no unit-test framework wired into CI**. Today:

- There is no `test` task in `mise.toml`, and `mise run build` (the CI gate) runs lint / fmt / validate / security scanners — but **no tests**.
- `pytest` (or any test runner) is not declared as a tool in `mise.toml` or `.pre-commit-config.yaml`.
- Exactly one test-shaped file exists — `tools/evals/databases-on-aws/dsql/scripts/test_safe_query.py` — and it is **not invoked** by `mise` or any workflow. It's a standalone/manual file.

Net effect: Python logic in `tools/` (and plugin scripts) can regress silently. Contributors have no standard place to put tests, and no assurance they run.

## Proposal

Establish a lightweight Python unit-test convention and gate it in CI:

1. **Add a test runner.** Add `pytest` to `mise.toml` `[tools]` (e.g. `"pipx:pytest"` or `"npm:"`-equivalent — pick per repo tooling norms). *(Adding a dependency — needs maintainer sign-off per the repo's "no new deps without asking" rule.)*
2. **Add a `mise run test` task** that discovers and runs tests (e.g. `pytest tools/ plugins/` or a curated set of paths).
3. **Wire it into `mise run build`** so CI fails on a broken test — add `{ task = "test" }` to the `build` chain in `mise.toml`.
4. **Document the convention** (test file location + naming, e.g. `test_*.py` next to the code or under a `tests/` dir) in `docs/DEVELOPMENT_GUIDE.md`.
5. **Handle bandit B101.** Bandit runs with defaults (no config), so `assert` in tests trips `B101`. Either add a bandit config that skips B101 for test paths, or follow the existing precedent of per-line `# nosec B101 - test assertion` markers (see `test_safe_query.py`, which carries 21 of them). A shared config is cleaner than per-line markers at scale.

## First example to add: guard the vendored-Semgrep parser (from #220)

Once the framework lands, add a test for `tools/semgrep/rules-update.py` (the vendored-ruleset updater introduced in #220). Its text-based `split_rules()` / `rule_id_of()` parser has a documented assumption worth guarding: **a rule's own content never begins a column-0 `- ` line** (that marker is what delimits rules). A reviewer flagged a fixture test as a good follow-up.

The test (already prototyped and passing during #220 review, then deferred to keep that PR scoped) should cover:

- **`split_rules` returns header + all blocks** and is **byte-preserving** (`header + "".join(blocks) == input`) — the property that keeps retained rules identical to the upstream download.
- **id extraction works when `id:` is not on the first line** — ~14 rules in `r/all` lead with `- patterns:` / `- fix:` and carry `id:` on a later line.
- **indented `- ` inside a `message`/`fix` block scalar does NOT cause a spurious split** (the positive side of the assumption).
- **a column-0 `- ` in content fails loud** — a mis-split fragment has no `id:`, so `rule_id_of` raises `UpdateError` (loud abort, not silent corruption). This is the fail-safe that makes the assumption tolerable.
- **`ID_RE` does not match nested keys** (`rule_id`, `r_id`, `rv_id`, `version_id`).
- **`split_rules` raises on a rule-less document** (`rules: []`).

Implementation notes for that test:
- `rules-update.py` has a hyphen, so import it via `importlib.util.spec_from_file_location` rather than a plain `import`.
- A zero-dependency `__main__` runner + pytest-shim (as in `test_safe_query.py`) lets it run via `uv run tools/semgrep/test_rules_update.py` even before the framework lands.

## Acceptance criteria

- [ ] `mise run test` exists and runs the Python test suite.
- [ ] `mise run build` fails if any test fails.
- [ ] Test convention documented in `docs/DEVELOPMENT_GUIDE.md`.
- [ ] Bandit B101-in-tests handled repo-wide (config or documented marker convention).
- [ ] The `rules-update.py` parser test above is included as the first example.

Follow-up to #220 (do not block that PR on this).

<sub>Filed with assistance from Claude Code.</sub>

scottschreckengaust:
Wires unit tests into the build so they run locally and in CI alike, using each language's standard test runner.

#### Related

- Closes #226 — establishes the test plumbing (a first test to *cover* the parser from #220 is a follow-up).

#### Problem

`mise run build` ran lint / fmt / validate / security but **no tests**. The one existing test file (`tools/evals/databases-on-aws/dsql/scripts/test_safe_query.py`) was never invoked, so Python logic under `tools/` could regress silently.

#### Approach

Use the ecosystem's standard runners rather than a bespoke script:

- **`test:py`** → `uv run --with pytest pytest tools plugins`. pytest already does discovery, per-file execution, and result aggregation. `--with` installs pytest into an ephemeral cached env for the run only, so **no dependency is added to the repo** — which was the whole reason the earlier custom shim existed. Exit code 5 (no tests collected) is treated as success.
- **`test:js`** → `npm test` only when a `package.json` defines a test script; otherwise skips. Leaves room for jest/other runners with no custom code.
- **`test`** aggregates the per-language tasks and stays in the build chain: `lint → fmt:check → validate → test → security`.

#### Why not a custom runner?

An earlier revision added a 91-line `tools/run-tests.py` that reimplemented discovery/aggregation/exclusion. pytest does all of that natively — and better: the custom runner's no-pytest fallback shim **silently skipped** parametrized and fixture-based tests. Switching to pytest raises the passing count from **63 → 65** while deleting ~91 lines.

#### Verification

- `mise run test` → `test:py` **65 passed**; `test:js` skips cleanly (no package.json).
- `mise run build` runs the `test` step; lint / fmt:check pass.
- Net change: **+18 / −94** (custom runner deleted).

#### Acknowledgment


theagenticguy (approved):
## ✅ Approve

Small, correct build-plumbing change (+27/−2 in `mise.toml` + `AGENTS.md`). Reviewed the full diff and verified the PR's claims against the committed head branch and CI config rather than the description alone.

**Why this is a solid approve:**

- **Correct exit-code handling.** `uv run --with pytest pytest tools plugins || [ $? -eq 5 ]` masks *only* pytest's exit 5 (no tests collected) as success. Real failures (exit 1) and collection/usage errors (2–4) still propagate and break the build.
- **No new repo dependency.** `--with pytest` installs into an ephemeral cached env for the run only — the whole reason the earlier custom shim existed, now unnecessary.
- **Standard runner over bespoke code.** pytest handles discovery/parametrize/fixtures natively; the deleted 91-line shim's fallback silently skipped parametrized/fixture tests (why the passing count rises 63→65).
- **CI is genuinely exercised.** The `build` job runs `mise run build`, so the new `test` step actually executes in CI (passed at 2m32s), and the mutation guard confirms the step writes nothing. All checks green.
- **Docs consistent.** Edits land in `AGENTS.md` (`CLAUDE.md` is a symlink to it). The updated line `lint + fmt:check + validate + test + security` also corrects a pre-existing omission — `validate` was already in the chain but undocumented.

Parser-covering tests are a documented fast-follow (#220), so their absence here is by design.

**Non-blocking nits** (fine to merge as-is):

1. The PR body names the tasks `test:py` / `test:js`, but the committed names are `test:python` / `test:node` — description drift only.
2. `test:node`'s `grep -q '"test"'` is a loose match against `package.json`; `jq -e '.scripts.test'` would be exact. Zero impact today (no `package.json`, so it always skips cleanly) — worth tightening only if a plugin ever ships one.

mayakost (approved):

PR #228, opened 2026-07-08
docs: document fast-forward-only rule for main

scottschreckengaust:
## What

Codifies the "only fast-forward `main`" rule in `AGENTS.md` (and via symlink, `CLAUDE.md`), so new worktrees and branches always start from the latest remote state.

## Changes (4 lines)

- **Worktree snippet**: fast-forward `main` (`git pull --ff-only origin main`) and branch new work off `origin/main`.
- **Boundary**: `ONLY fast-forward main (git pull --ff-only); never commit, merge, or rebase it locally.`

Kept intentionally terse — modern agents already understand fast-forward semantics; this just states the project's expectation.

## Verification

- `mise run lint:md` → 0 errors.
- Edited `AGENTS.md` only; `CLAUDE.md` inherits via symlink.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -174,7 +174,7 @@ claude --plugin-dir ./plugins/deploy-on-aws
 
 ## Git Worktree Workflow
 
-ALWAYS use git worktrees for new work. The main worktree stays on its current branch and is never switched. Each piece of work gets its own worktree under `.tmp/`, branching off the current branch. This enables multiple agents to work in parallel without conflicts.
+ALWAYS use git worktrees for new work. The main worktree stays on its current branch and is never switched. Each piece of work gets its own worktree under `.tmp/`, branching off the latest `origin/main`. This enables multiple agents to work in parallel without conflicts.
 
 ```bash
 # Fast-forward main first, then branch new work off the latest remote
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -177,8 +177,9 @@ claude --plugin-dir ./plugins/deploy-on-aws
 ALWAYS use git worktrees for new work. The main worktree stays on its current branch and is never switched. Each piece of work gets its own worktree under `.tmp/`, branching off the current branch. This enables multiple agents to work in parallel without conflicts.
 
 ```bash
-# Create a worktree for new work (branches off current branch)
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -177,8 +177,9 @@ claude --plugin-dir ./plugins/deploy-on-aws
 ALWAYS use git worktrees for new work. The main worktree stays on its current branch and is never switched. Each piece of work gets its own worktree under `.tmp/`, branching off the current branch. This enables multiple agents to work in parallel without conflicts.
```
theagenticguy: Nit (non-blocking): this line still says *"branching off the current branch,"* but the command just below now branches off `origin/main` (`git worktree add .tmp/<short-name> -b <branch-name> origin/main`). After a fresh `git pull --ff-only` the two are the same, so it's harmless — but the prose and the command now say slightly different things. Suggest syncing the sentence, e.g. "...branching off the latest `origin/main`.", so the narrative matches the code.

theagenticguy (commented):
Reviewed the full diff plus the surrounding **Git Worktree Workflow** and **Boundaries** sections. The change is correct and CI is green (23/23).

- `git pull --ff-only origin main` followed by `git worktree add .tmp/<name> -b <branch> origin/main` correctly branches new work off the latest remote, and `--ff-only` self-enforces the new boundary — a stray local commit on `main` makes the pull fail loudly rather than silently diverging.
- Confirmed `CLAUDE.md` is a symlink to `AGENTS.md` (git mode `120000`), so editing `AGENTS.md` alone is correct.

One optional prose nit inline. Posting as **Comment** (not Approve) per request — happy to leave the recommended fix to you.

_Reviewed by Bonk._

AGENTS.md
scottschreckengaust: fixed

theagenticguy (approved):
Re-reviewed after the prose fix. Line 177 now reads "branching off the latest `origin/main`," so the narrative and the command are in sync. The rest of the diff is unchanged and correct, and CI is green on `bc96785` (all checks pass).

Approving. Correct, coherent, and terse-by-design. Thanks for the quick turnaround on the nit.

_Reviewed by Bonk._

krokoko (approved):

PR #237, opened 2026-07-21
fix(hyperpod-nccl): make nccl-diagnose.sh compatible with bash 3.2

ssuday:
## What

Makes `plugins/sagemaker-ai/skills/hyperpod-nccl/scripts/nccl-diagnose.sh` compatible with **bash 3.2** (the default `/bin/bash` on macOS and some minimal/older environments). The script previously used bash 4.0+ only constructs and would fail to run there.

## Changes

Three bash 4.0+ features replaced with portable equivalents — no behavior change:

1. **`${VAR^^}` uppercase expansion** (bash 4.0+) → `printf '%s' "$VAR" | tr '[:lower:]' '[:upper:]'`
2. **`declare -A` associative array** for the NCCL log-pattern table (bash 4.0+) → two parallel indexed arrays (`NCCL_PAT_KEYS` / `NCCL_PAT_VALS`) iterated by index
3. **heredoc into command substitution** for the on-node diagnostic script → `IFS= read -r -d '' _NODE_DIAG_SCRIPT <<'EOF'`

## Testing

Ran the full diagnostic **live against a real HyperPod EKS cluster** (2× `ml.p4d.24xlarge`, A100) under both:

- **GNU bash 4.2.46** (Amazon Linux default)
- **GNU bash 3.2.57** (built from GNU source to get a genuine glibc 3.2 runtime)

Both runs completed end-to-end — cluster health, security groups, all K8s checks, NCCL log-pattern analysis (correctly detected `DNS_FAIL` / `CONN_REFUSED`), and on-node SSM hardware checks — and produced **structurally identical output** (same PASS/FAIL/WARN findings) with the **same exit code**. `bash -n` also passes under the `bash:3.2` image.

### DIFF
--- a/plugins/sagemaker-ai/skills/hyperpod-nccl/scripts/nccl-diagnose.sh
+++ b/plugins/sagemaker-ai/skills/hyperpod-nccl/scripts/nccl-diagnose.sh
@@ -226,7 +226,7 @@ detect_orchestrator() {
         warn "Could not auto-detect orchestrator — defaulting to 'eks'"
         warn "Override with: --orchestrator slurm"
     fi
-    success "Orchestrator: ${ORCHESTRATOR^^}"
+    success "Orchestrator: $(printf '%s' "$ORCHESTRATOR" | tr '[:lower:]' '[:upper:]')"
 }
 
 check_prerequisites() {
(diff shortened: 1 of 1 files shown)


badmin-aws (approved):

singh96aman (approved):
The changes are correct and well-tested. Consider adding the array length guard to prevent future maintenance issues. 

The parallel `NCCL_PAT_KEYS` / `NCCL_PAT_VALS` arrays must stay perfectly synchronized. A future edit could easily add a key without its value (or vice versa), causing silent mismatches.

Suggestion: Add a runtime sanity check after the array definitions:

```
if [[ ${#NCCL_PAT_KEYS[@]} -ne ${#NCCL_PAT_VALS[@]} ]]; then
    error "Internal error: NCCL_PAT_KEYS/VALS length mismatch"
    return 1
fi
```

This catches maintenance mistakes early.

scottschreckengaust (approved):
LGTM

Issue #148, opened 2026-04-24
RFC: AWS Containers plugin


niallthomson:
### Summary

Add a plugin to help developers containerize, deploy, manage, and troubleshoot containerized applications on AWS using Amazon ECS, Amazon EKS, and related container services (ECR). The plugin provides a unified entry point for container workloads on AWS, guiding developers from Dockerfile creation through production operations.

This plugin will also serve as a mechanism to package skills for open source projects which are maintained by AWS but are not strictly related to services. Examples of these include:

1. [Karpenter](https://karpenter.sh/)
1. [AWS Controllers for Kubernetes](https://aws-controllers-k8s.github.io/docs/) (ACK)
1. [Kube Resource Orchestration](https://kro.run/) (kro)

### Use case

These are example use cases and is not an exhaustive list:

- Containerization of existing applications: Generating production-ready Dockerfiles with multi-stage builds, selecting appropriate base images, configuring .dockerignore files, building images, pushing to ECR and handling ECR lifecycle policies. Covers language-specific patterns (e.g., dependency caching for Node.js/Python/Java), image size optimization, building images compatible with Graviton and security hardening (non-root users, minimal base images, vulnerability scanning).
- Troubleshooting common container issues: Diagnosing why containers fail to start, crash-loop, or misbehave in ECS/EKS environments. Covers signal handling (catching SIGTERM for graceful shutdown), health check configuration and failure diagnosis, image pull errors (ECR auth, missing tags, cross-account access), OOMKilled pods/tasks from incorrect memory limits, networking issues (security groups, service discovery, load balancer target registration), and reading CloudWatch/Container Insights logs to trace root cause.
- ECS Express Mode deployments for rapid prototyping: Going from application code to a running ECS service with a public URL in minutes. Express Mode auto-provisions infrastructure (VPC, ALB, ECR, task definitions, service) so developers can validate their containerized app on AWS without writing IaC. Covers the full Express Mode lifecycle -- containerization, image build/push, prerequisite validation, deployment, readiness polling, and cleanup.
[...]

Issue #55, opened 2026-02-24
RFC: Amazon Location Service plugin


cgalvan:
### Summary

This RFC proposes adding an `amazon-location-service` plugin to the agent-plugins-for-aws marketplace. The plugin provides an agent skill that guides developers through integrating geospatial features into their applications using Amazon Location Service APIs, covering maps, places search, geocoding, routing, address verification, and authentication setup.

### Use case

Developers building location-aware applications on AWS currently need to navigate multiple API references, SDK documentation, and best practices guides across Amazon Location Service's five service areas (Places, Maps, Routes, Geofences, and Trackers). Common pitfalls include using wrong API action names in permissions, incorrect map initialization patterns, and choosing suboptimal APIs for their use case.

This plugin encodes that domain expertise into a skill that auto-triggers when developers express location-related intent (e.g., "add a map", "geocode an address", "calculate a route"). It provides API selection guidance, authentication setup, SDK integration patterns, and reference documentation for specific implementation tasks.

### Proposal

The plugin follows the established plugin structure:

```
plugins/amazon-location-service/
├── .claude-plugin/
│   └── plugin.json
├── .mcp.json                          # aws-mcp server for documentation access
└── skills/
    └── amazon-location-service/
        ├── SKILL.md                   # Core skill with API guidance and common mistakes
        └── references/
            ├── address-input.md       # Address autocomplete form patterns
            ├── address-verification.md # Address validation before persistence
            ├── calculate-routes.md    # Route calculation and map display
            ├── dynamic-map.md         # MapLibre integration with Amazon Location tiles
            ├── places-search.md       # Text search, nearby search, suggestions
            └── web-javascript.md      # Browser SDK integration patterns
```

The skill covers:
- API selection guidance (which API to use for which use case)
- Common mistakes and how to avoid them (wrong display fields, incorrect map init, mixed resource types)
- Authentication patterns (API keys for client-side, IAM for server-side)
- Preference for resourceless operations over resource-based ones
- Preference for the bundled JavaScript client (@aws/amazon-location-client) for web development
- Reference files for progressive disclosure of implementation details

The plugin integrates with the AWS MCP Server for access to live documentation and API references.

Changes outside the plugin directory include registering it in `marketplace.json`, updating `README.md`, `AGENTS.md`, `MAINTAINERS_GUIDE.md`, `DESIGN_GUIDELINES.md`, and `.github/copilot-instructions.md` to reflect the new plugin.

### Out of scope

[...]

PR #243, opened 2026-07-25
feat(aws-agent-payments): add x402 paywall auto-pay plugin for autonomous agents

wirjo:
## Summary

Adds a new `aws-agent-payments` plugin that enables stateful autonomous agents to transparently pay for x402-paywalled APIs, MCP tools, and web content via microtransactions using [AWS AgentCore Payments](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/payments.html).

## Related

- [x402 Protocol](https://www.x402.org/) — open HTTP payment standard
- [AgentCore Payments docs](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/payments.html)
- End-to-end validated with Coinbase CDP on Base Sepolia against `x402-test.genesisblock.ai`

### New Plugin: `plugins/aws-agent-payments/`

```
plugins/aws-agent-payments/
├── .claude-plugin/plugin.json          # Claude Code marketplace
├── .codex-plugin/plugin.json           # Codex marketplace
├── .mcp.json                           # AWS MCP Server config
├── README.md                           # Full documentation
├── openclaw.plugin.json                # Runtime plugin manifest (NEW PATTERN)
├── package.json / tsconfig.json
├── src/                                # TypeScript reference implementation (NEW PATTERN)
│   ├── index.ts                        # Plugin entry + tool registration
│   ├── config.ts                       # Config management
│   ├── payments.ts                     # AgentCore API integration
│   └── x402.ts                         # x402 protocol handling (v1 + v2)
└── skills/x402-payments/
    ├── SKILL.md                        # Agent-agnostic skill instructions
    └── references/
        ├── protocol.md                 # x402 protocol details (v1/v2)
        ├── setup.md                    # Prerequisites + IAM setup
        └── debugging.md               # Common issues + diagnostics
```

### New Standards Introduced

This plugin introduces two new patterns that extend the existing plugin model:

#### 1. Runtime Plugin Manifest (`openclaw.plugin.json`)

Unlike skill-based plugins (which provide markdown instructions for the agent to follow), this manifest declares **executable tools** that stateful agent hosts register directly into their tool system:

```json
{
  "id": "aws-agent-payments",
  "name": "AWS Agent Payments (x402)",
  "contracts": {
    "tools": [
      "get_payment_session_status",
      "create_payment_session",
      "get_paid_content",
      "pay_and_get_header",
      "setup_x402_payments"
    ]
  },
  "activation": { "onStartup": true },
  "configSchema": { ... }
}
```

This enables fully autonomous payment flows without human-in-the-loop for each transaction — critical for agents that browse, fetch APIs, and act independently.

#### 2. TypeScript Reference Implementation (`src/`)

While all existing plugins are skill-only (markdown + optional scripts), this plugin includes a working TypeScript implementation that demonstrates:

- x402 challenge parsing (v1 and v2 protocol versions)
- AgentCore `ProcessPayment` integration via AWS SDK
- Payment-Signature envelope construction
- Session lifecycle management (creation, validation, expiry)
- Full infrastructure setup automation

This serves as a **reference implementation** that other agent hosts can adapt. The SKILL.md continues to work independently for coding agents following the existing pattern.

### Why These Patterns Matter

Stateful autonomous agents (OpenClaw, Hermes, custom AgentCore deployments) differ from coding agents in a key way: they act **continuously and autonomously** — browsing the web, calling APIs, managing infrastructure — without a developer in the loop. For x402 payments, this means:

- The agent must handle 402s **in real-time** without asking the user to run a script
- Payment sessions must be managed **across conversations** (persist state)
- Tools must be **always available** (activated on startup, not invoked via slash command)

Skill-only plugins work well for coding agents (developer reads guidance, runs CLI commands). Runtime plugins work for autonomous agents (tools execute programmatically, no human steps).

### Tools Provided

| Tool | Description |
|---|---|
| `setup_x402_payments` | One-shot infrastructure setup (Payment Manager, Connector, Wallet) |
| `get_payment_session_status` | Check current session usability |
| `create_payment_session` | Mint fresh session (locks USDC up to spend cap) |
| `get_paid_content` | Pay + fetch content server-side (recommended) |
| `pay_and_get_header` | Mint payment header for browser-based flows |

### Supported Providers & Networks

- **Wallet providers:** Coinbase CDP, Stripe/Privy
- **Networks:** Base Sepolia (testnet), Base Mainnet, Solana Devnet, Solana Mainnet

## Future: ClawHub Publishing Lifecycle

The runtime plugin pattern introduced here is designed to support publishing via [ClawHub](https://clawhub.ai) — the OpenClaw skill and plugin registry. This would enable one-command install and automated lifecycle management:

```bash
# End users install via OpenClaw CLI:
openclaw plugins install clawhub:@aws/aws-agent-payments

# Publishers manage versions via clawhub CLI:
clawhub login
clawhub package publish . --family code-plugin --dry-run
clawhub package publish . --family code-plugin
```

**Implementation path:**

1. **Compatibility metadata** — add `openclaw.compat.pluginApi` and `openclaw.build.openclawVersion` to `package.json` so ClawHub can validate version compatibility at install time
2. **Build pipeline** — `openclaw plugins build` generates the final manifest from TypeScript entry metadata, keeping `openclaw.plugin.json` and `package.json` in sync
3. **CI auto-publish** — a GitHub Action publishes new versions to ClawHub on tagged releases, giving users automatic updates
4. **Skill publishing** (independent of plugin) — `clawhub skill publish ./skills/x402-payments` makes the SKILL.md discoverable for any agent host, not just OpenClaw

This gives a full lifecycle: develop locally → validate → publish to registry → users install with one command → auto-update on new releases. The same pattern could be adopted by other plugins in this repository that want to support autonomous agent hosts.

## Acknowledgment

### DIFF
--- a/plugins/aws-agent-payments/.claude-plugin/plugin.json
+++ b/plugins/aws-agent-payments/.claude-plugin/plugin.json
@@ -1,0 +1,28 @@
+{
+  "author": {
+    "name": "Amazon Web Services"
+  },
+  "description": "Enable stateful autonomous agents to pay for x402-paywalled APIs, MCP tools, and web content via microtransactions using AWS AgentCore Payments. Supports Coinbase CDP and Stripe/Privy wallet providers.",
+  "homepage": "https://github.com/awslabs/agent-plugins",
+  "keywords": [
+    "aws",
+    "payments",
+    "x402",
+    "agentcore",
+    "bedrock-agentcore",
+    "micropayments",
+    "usdc",
+    "crypto",
+    "paywall",
+    "coinbase",
+    "stripe",
+    "privy",
+    "base",
+    "web3",
+    "autonomous-agents"
+  ],
+  "license": "Apache-2.0",
+  "name": "aws-agent-payments",
+  "repository": "https://github.com/awslabs/agent-plugins",
+  "version": "1.0.0"
+}
--- a/plugins/aws-agent-payments/.codex-plugin/plugin.json
+++ b/plugins/aws-agent-payments/.codex-plugin/plugin.json
@@ -1,0 +1,43 @@
+{
+  "name": "aws-agent-payments",
+  "version": "1.0.0",
+  "description": "Enable stateful autonomous agents to pay for x402-paywalled APIs, MCP tools, and web content via microtransactions using AWS AgentCore Payments. Supports Coinbase CDP and Stripe/Privy wallet providers.",
+  "author": {
+    "name": "Amazon Web Services",
+    "email": "aws-agent-plugins@amazon.com",
+    "url": "https://github.com/awslabs/agent-plugins"
+  },
+  "homepage": "https://github.com/awslabs/agent-plugins",
+  "repository": "https://github.com/awslabs/agent-plugins",
+  "license": "Apache-2.0",
+  "keywords": [
+    "aws",
+    "payments",
+    "x402",
+    "agentcore",
+    "micropayments",
+    "usdc",
+    "autonomous-agents"
+  ],
+  "skills": "./skills/",
+  "interface": {
+    "displayName": "AWS Agent Payments (x402)",
+    "shortDescription": "Pay for x402-paywalled APIs and content via AgentCore Payments.",
+    "longDescription": "Enable agents to transparently pay for x402-paywalled APIs, MCP tools, and web content via microtransactions using AWS AgentCore Payments.",
+    "defaultPrompt": [
+      "Pay for this x402-paywalled URL.",
+      "Set up x402 payments with my Coinbase CDP credentials.",
+      "Check my payment session status."
+    ],
+    "developerName": "Amazon Web Services",
+    "category": "Payments",
+    "capabilities": [
+      "Read",
+      "Write"
+    ],
+    "websiteURL": "https://github.com/awslabs/agent-plugins",
+    "privacyPolicyURL": "https://aws.amazon.com/privacy/",
+    "termsOfServiceURL": "https://aws.amazon.com/service-terms/",
+    "brandColor": "#FF9900"
+  }
+}
--- a/plugins/aws-agent-payments/README.md
+++ b/plugins/aws-agent-payments/README.md
@@ -1,0 +1,82 @@
+# Agent Payments (x402)
+
+> 🚧 **Preview:** AWS AgentCore Payments is currently in **preview**. APIs, pricing, and availability may change. See [AgentCore Payments documentation](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/payments.html) for the latest status.
+
+Enable stateful autonomous agents to pay for x402-paywalled APIs, MCP tools, and web content via microtransactions using [AWS AgentCore Payments](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/payments.html).
+
+## Overview
+
+When an AI agent encounters an HTTP 402 (Payment Required) response from an x402-protected endpoint, this skill guides the agent through the payment flow — detecting the challenge, processing payment via AgentCore, and replaying the request with a valid payment header. The agent receives the content without needing to understand the underlying crypto mechanics.
+
+Works with stateful autonomous agents (OpenClaw, custom AgentCore deployments) and AI coding agents that need to access paid APIs or content.
+
+Supports **Coinbase CDP** and **Stripe/Privy** wallet providers on **Base Sepolia** (testnet) and **Base Mainnet** networks.
+
+## Skills
+
+| Skill           | When to use                                                              | References                                                                                                                                                            |
+| --------------- | ------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `x402-payments` | "pay for this URL", "x402 paywall", HTTP 402 detected, "set up payments" | [protocol](skills/x402-payments/references/protocol.md), [setup](skills/x402-payments/references/setup.md), [debugging](skills/x402-payments/references/debugging.md) |
+
+## How x402 Payment Works
+
+```text
+Agent request → HTTP 402 + x402 challenge
+    → AgentCore ProcessPayment (signs tx)
+    → Replay request with X-PAYMENT header
+    → HTTP 200 + paid content returned to agent
+```
+
+## Installation
+
+### Claude Code
+
+```bash
+/plugin marketplace add awslabs/agent-plugins
+/plugin install aws-agent-payments@agent-plugins-for-aws
+```
+
+### Codex
+
+```bash
+codex plugin marketplace add awslabs/agent-plugins
+```
+
+Then install **aws-agent-payments** from the Plugins panel.
+
+## Prerequisites
+
+- AWS account with [AgentCore Payments](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/payments.html) access
+- AWS CLI with configured credentials (`aws sts get-caller-identity`)
+- IAM role with `bedrock-agentcore.amazonaws.com` trust policy
+- Wallet provider account:
+  - **Coinbase CDP**: API key from [portal.cdp.coinbase.com](https://portal.cdp.coinbase.com)
+  - **Stripe/Privy**: App credentials from [dashboard.privy.io](https://dashboard.privy.io)
+- USDC funding on the target network (Base Sepolia for testnet)
+
+## Security & Spending Controls
+
+- **Session spend caps** — each session locks a maximum USDC amount (default $5)
+- **Session TTL** — sessions auto-expire (default 4 hours, max 8 hours)
+- **Per-request limits** — AgentCore enforces per-transaction caps from the x402 challenge
+- **IAM policies** — scope agent permissions to only `ProcessPayment`, `GetPaymentSession`
+- **Audit trail** — all payments logged via CloudTrail
+
+## Supported Networks
+
+| Network      | Chain ID       | Use Case              |
+| ------------ | -------------- | --------------------- |
+| Base Sepolia | `eip155:84532` | Testnet / development |
+| Base Mainnet | `eip155:8453`  | Production            |
+
+## Examples
+
+- "Pay for this API endpoint: `https://x402-test.example.com/api/weather`"
+- "Set up x402 payments with my Coinbase CDP credentials"
+- "Check my payment session balance"
+- "Create a new payment session with $10 cap"
+- "This URL returned a 402, can you pay for it?"
+
+## Runtime Plugin (Future)
+
+A TypeScript reference implementation for stateful agent hosts (OpenClaw, etc.) that registers executable tools directly into the agent's tool system is available at [wirjo/agent-toolkit-for-aws](https://github.com/wirjo/agent-toolkit-for-aws/pull/1). This enables fully autonomous payment flows without human-in-the-loop for each transaction. An RFC for the runtime-plugin pattern will be opened separately.
--- a/plugins/aws-agent-payments/skills/x402-payments/SKILL.md
+++ b/plugins/aws-agent-payments/skills/x402-payments/SKILL.md
@@ -1,0 +1,183 @@
+---
+name: x402-payments
+description: "Auto-pay x402-paywalled URLs using AWS AgentCore Payments. Handles HTTP 402 challenges, payment signing, and request replay for both v1 and v2 of the x402 protocol."
+metadata:
+  tags: [x402, payments, paywall, usdc, crypto, web3, agentcore, micropayments]
+  version: "1.0.0"
+---
+
+# x402 Payments
+
+This skill enables AI agents to auto-pay x402-paywalled URLs using AWS AgentCore Payments. It is agent-host independent — it describes the protocol and tool interactions generically, working the same whether tools are exposed as direct functions, MCP tools, or namespaced actions.
+
+## Tool Inventory
+
+The x402 payment system provides five tools. Match them by role — your runtime may prefix or rename them:
+
+| Role                    | Typical name                 | Purpose                                                   |
+| ----------------------- | ---------------------------- | --------------------------------------------------------- |
+| Infrastructure setup    | `setup_x402_payments`        | One-shot creation of Payment Manager, Connector, Wallet   |
+| Session status check    | `get_payment_session_status` | Check if current session is usable                        |
+| Session creation        | `create_payment_session`     | Mint a fresh session with budget (requires user approval) |
+| Server-side pay + fetch | `get_paid_content`           | Pay and return content in one call                        |
+| Header-only payment     | `pay_and_get_header`         | Mint payment header for browser replay                    |
+
+## When to Use
+
+Activate this skill when **any** of these occur:
+
+- An HTTP request returns 402 with x402 challenge headers (`x402Version`, `Payment-Required`, or `x-payment-required`)
+- The user asks to access content from a URL they identify as x402-paywalled
+- The user asks to "set up x402 payments" or "configure agent payments"
+- A tool call fails with 402 Payment Required
+
+Do NOT use for:
+
+- Login walls, captchas, or non-x402 paywalls
+- AWS billing or cost management
+- Stripe checkout / e-commerce payments
+
+## Protocol
+
+### Step 1: Check Payment Session
+
+Call the **session-status** tool. The response includes a `usable` boolean.
+
+- **If `usable: true`** → proceed to Step 3
+- **If `usable: false`** → the session is expired, drained, or doesn't exist. Go to Step 2.
+
+### Step 2: Request User Approval for New Session
+
+**Never mint a session without explicit user approval.** Each session sets a spending budget that the agent can use.
+
+Tell the user:
+
+> Your payment session is [expired/drained/missing]. I can create a new one with a $5 budget valid for 4 hours. Would you like to approve this?
+
+On approval, call the **create-session** tool with `max_spend_usd="5"` and `expiry_minutes=240` (or user-specified values). Always ask for budget and duration confirmation before proceeding.
+
+### Step 3: Pay for the URL
+
+**Preferred path (server-side):** Call `get_paid_content` with the URL. The tool:
+
+1. Probes the URL → gets 402 + x402 challenge
+2. Calls AgentCore ProcessPayment → gets signed payment header
+3. Replays the request with the payment header
+4. Returns `{status_code, content_type, body, url}`
+
+Read the content from the `body` field. No second request needed.
+
+**Alternative path (browser/header-only):** If you need the page rendered in a live browser:
+
+1. Call `pay_and_get_header` → returns `{header: {"X-PAYMENT": "VALUE"}, valid_seconds: <derived from challenge>}`
+2. Set the header on your browser context
+3. Navigate to the URL again — replay immediately, header validity is time-limited
+4. The paid page renders normally
+
+### Step 4: Handle Errors
+
+- **Session expired mid-request** → go back to Step 2
+- **Still 402 after payment** → session may have drained; check session status before retrying, then mint a new session if balance is zero
+- **Header expired** → call `pay_and_get_header` again
+
+For detailed error diagnosis, see [references/debugging.md](references/debugging.md).
+
+## First-Time Setup
+
+If no payment infrastructure exists, use the **setup** tool. For detailed prerequisites and IAM configuration, see [references/setup.md](references/setup.md).
+
+Required inputs:
+
+- `role_arn` — IAM role with `bedrock-agentcore.amazonaws.com` trust policy
+- Wallet provider credentials (stored securely in AWS Secrets Manager, never in transcripts)
+- Optional: `region` (default us-east-1), `network` (default eip155:84532), `user_id`
+
+After setup:
+
+1. Provide the redirect URL for delegated signing authorization
+2. Instruct the user to fund the wallet with USDC on the target network
+
+## x402 Protocol Details
+
+For the full protocol specification including v2 envelope format, supported assets, and chain IDs, see [references/protocol.md](references/protocol.md).
+
+## Guidelines
+
+- **Don't surface payment internals** — report the content, not transaction hashes or header bytes
+- **Always check session status first** — avoids `ExpiredTokenException` errors
+- **Never auto-mint sessions** — get explicit user approval each time (unless user previously authorized auto-creation)
+- **Wallet credentials must never appear in tool parameters or transcripts** — read from environment or secure config at execution time
+- **Verify session balance before retrying** — a retry after failure should confirm the session still has funds rather than blindly calling ProcessPayment again
+
+## Session Consent Enforcement
+
+`CreatePaymentSession` is the spending-authorization boundary — whoever calls it controls how much the agent can spend. If the agent can create sessions autonomously, it can bypass budget controls by minting new sessions indefinitely.
+
+**Recommended enforcement (strongest → weakest):**
+
+| Tier | Mechanism | Guarantee |
+|------|-----------|----------|
+| 1 | **Host-native approval gate** — the agent host intercepts `CreatePaymentSession` and surfaces an Approve/Deny prompt to the user. The LLM cannot bypass this. | Hard |
+| 2 | **IAM role separation** — the agent's runtime role only has `ProcessPayment` + `GetPaymentSession`. Sessions are created out-of-band by a human or privileged process. | Hard |
+| 3 | **Conversational confirmation** — the agent asks the user "Create a $5 session?" and waits for affirmative response. | Soft (LLM-enforceable only) |
+
+**Recommendations:**
+
+- For **unattended/autonomous agents**: use Tier 1 or Tier 2. Tier 3 alone is insufficient — a jailbroken or confused model could self-approve.
+- For **interactive chat agents with a human present**: Tier 1 (host approval gate) is ideal. Tier 3 is acceptable as a fallback if the host lacks native approval.
+- For **developer/CLI workflows**: session creation should be a human-run script outside the LLM loop entirely (Tier 2).
+
+**Host-specific implementations:**
+
+- **OpenClaw**: `create_payment_session` uses a two-phase confirmation gate. The first call returns `AWAITING_USER_APPROVAL` and the agent must present the budget/duration to the user. Only after explicit approval does the second call (with `confirmed: true`) execute.
+- **Strands/LangGraph**: Use IAM role separation (Tier 2). The `setup_payment_user.py` script creates sessions in the terminal; the agent runtime role lacks `CreatePaymentSession`.
+- **Custom hosts**: Implement a pre-execution hook on `CreatePaymentSession` that gates on user confirmation via your application's consent mechanism.
+
+## Fallback (No Runtime Plugin)
+
+If runtime tools (`get_paid_content`, `pay_and_get_header`) are not available, use the AWS CLI directly:
+
+### Check/Create Session
+
+```bash
+# Check existing session
+aws bedrock-agentcore get-payment-session \
+  --payment-manager-arn $PM_ARN \
+  --payment-session-id $SESSION_ID \
+  --user-id $USER_ID \
+  --region us-east-1
+
+# Create new session (requires user approval for budget)
+aws bedrock-agentcore create-payment-session \
+  --payment-manager-arn $PM_ARN \
+  --user-id $USER_ID \
+  --budget '{"maxSpendUsd": "5.00"}' \
+  --ttl-minutes 240 \
+  --region us-east-1
+```
+
+### Process Payment
+
+```bash
+# After receiving a 402 with x402 challenge, extract the accepts[0] object and call:
+aws bedrock-agentcore process-payment \
+  --payment-manager-arn $PM_ARN \
+  --payment-session-id $SESSION_ID \
+  --payment-instrument-id $INSTRUMENT_ID \
+  --user-id $USER_ID \
+  --payment-type CRYPTO_X402 \
+  --payment-input '{"cryptoX402": {"version": "2", "payload": <accepts[0] object>}}' \
+  --region us-east-1
+```
+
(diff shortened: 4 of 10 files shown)


plugins/aws-agent-payments/.codex-plugin/plugin.json
```diff
@@ -0,0 +1,28 @@
...
+    "crypto",
+    "paywall",
+    "coinbase",
+    "stripe",
+    "privy",
+    "base",
+    "web3",
+    "autonomous-agents"
+  ],
+  "license": "Apache-2.0",
```
theagenticguy: **Blocker — build gate fails.** `mise run lint:manifests` rejects this file:

```
plugins/aws-agent-payments/.codex-plugin/plugin.json invalid
data must have required property 'interface'
```

`schemas/codex-plugin.schema.json` requires `interface` with `displayName`, `shortDescription`, `longDescription`, `developerName`, `category`, and `capabilities` (see `plugins/aws-amplify/.codex-plugin/plugin.json` for the shape). The other nine plugins all pass.

Note this file is generated, not hand-written: `tools/generate_codex_manifests.py` builds it from `.claude-plugin/marketplace.json` plus an `INTERFACE_METADATA` entry keyed on plugin name. Since `aws-agent-payments` is absent from both, the generator would skip it entirely. Fix in this order: add the marketplace entry, add the `INTERFACE_METADATA` block, then regenerate rather than editing by hand.

plugins/aws-agent-payments/README.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+## Installation
+
+### Claude Code
+
+```bash
+/plugin marketplace add awslabs/agent-plugins
+/plugin install aws-agent-payments@agent-plugins-for-aws
+```
```
theagenticguy, line 37: **Blocker — plugin is not registered in either marketplace.** `.claude-plugin/marketplace.json` lists eight plugins and `.agents/plugins/marketplace.json` lists nine; `aws-agent-payments` is in neither.

That makes the documented install commands here non-functional: `/plugin install aws-agent-payments@agent-plugins-for-aws` resolves through the marketplace manifest, and there is no entry to resolve. `mise run lint:cross-refs` passes only because it validates marketplace-to-directory, never the reverse, so an unregistered plugin directory is invisible to it.

plugins/aws-agent-payments/skills/x402-payments/SKILL.md
```diff
@@ -0,0 +1,110 @@
+---
+name: x402-payments
+description: >-
```
theagenticguy: **Blocker — `mise run lint:md` fails here (exit 1).** Two frontmatter errors:

```
SKILL.md:13 skill-frontmatter/SKILL002 [Description should be at least 20 characters (current: 2)]
SKILL.md:13 skill-frontmatter/SKILL002 [Non-spec frontmatter property: "version"]
```

The description reads as 2 characters because `tools/markdownlint-frontmatter.cjs` matches multiline descriptions with `/^description:\s*>\s*\n/` — a bare `>`, not `>-`. The strip indicator makes the multiline branch miss, the single-line branch then captures the literal `>-`, and the length check fails on that. Every other SKILL.md in the repo uses a quoted single-line description, which sidesteps this. Either follow that convention or fix the rule's regex.

Separately, `version` is not in the Agent Skills spec allowlist (`agent`, `allowed-tools`, `argument-hint`, `context`, `description`, `disable-model-invocation`, `license`, `metadata`, `model`, `name`, `user-invocable`). Move it under `metadata` if you want to keep it.

plugins/aws-agent-payments/skills/x402-payments/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+1. Provide the redirect URL for delegated signing authorization
+2. Instruct the user to fund the wallet with USDC on the target network
+
+## Guidelines
+
+- **Don't surface payment internals** — report the content, not transaction hashes or header bytes
+- **Always check session status first** — avoids `ExpiredTokenException` errors
+- **Never auto-mint sessions** — get explicit user approval each time (unless user previously authorized auto-creation)
+- **Payment headers expire in ~60 seconds** — if using the header path, replay immediately
+- **Idempotent retries** — ProcessPayment uses a client_token for idempotency, safe to retry on transient failures
```
theagenticguy: **Money-loss risk — this idempotency guarantee does not hold.** The line tells the agent retries are safe because "ProcessPayment uses a client_token for idempotency."

`clientToken` on `ProcessPaymentRequest` carries the Smithy `idempotencyToken` trait, and `src/payments.ts` never sets it. The SDK therefore auto-fills a **fresh UUID on every call** (`@smithy/core/dist-cjs/submodules/protocols/index.js:836` — `if (ns.isIdempotencyToken()) this.stringBuffer = generateIdempotencyToken()`). Two calls for the same 402 get two different tokens and are two distinct payments.

So the exact scenario this guidance encourages is the one that double-spends: `ProcessPayment` succeeds server-side, the response times out, the agent retries on the "safe to retry" advice, and the user pays twice. Either thread a caller-stable `clientToken` derived from the challenge (nonce/resource/amount) so a retry is genuinely deduplicated, or delete the claim and tell the agent to verify session state before retrying.

plugins/aws-agent-payments/src/payments.ts
```diff
@@ -0,0 +1,351 @@
...
+  // The payload is a DocumentType — normalize to object
+  const signedPayloadObj: Record<string, unknown> =
+    typeof signedPayload === "string" ? JSON.parse(signedPayload) : signedPayload;
+
+  // For v2: header is Payment-Signature; for v1: X-PAYMENT
+  const headerName = version === "2" ? "Payment-Signature" : "X-PAYMENT";
+
+  return {
+    headerName,
+    headerValue: JSON.stringify(signedPayloadObj), // raw JSON; caller builds envelope
```
theagenticguy: **Protocol bug — the v1 payment header is neither base64-encoded nor wrapped in the required envelope.** For `version !== "2"` this raw JSON string is passed straight through to the `X-PAYMENT` header (`src/index.ts:130`, `:205`).

Two deviations from the spec:

1. **Encoding.** `specs/transports-v1/http.md` is explicit: "**Mechanism**: `X-PAYMENT` header containing base64-encoded JSON." Raw JSON also risks a header-parsing failure on the non-ASCII/whitespace path.
2. **Shape.** `specs/x402-specification-v1.md` §5.2 defines the v1 `PaymentPayload` as `{x402Version, scheme, network, payload}` — where the signed `{signature, authorization}` object goes *inside* `payload`. This sends the inner object alone, so `x402Version`, `scheme`, and `network` are all missing.

Result: any real v1 server rejects the header. The PR notes end-to-end validation against `x402-test.genesisblock.ai`, which suggests only the v2 path was exercised. If v1 has not been tested against a live server, consider dropping it from this PR rather than shipping an untested branch, since the README and `references/protocol.md` both advertise it.

plugins/aws-agent-payments/src/x402.ts
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Build the ProcessPayment payload from the accepted challenge entry.
+ * AgentCore expects the accepts object fields (scheme, network, amount, asset, payTo, etc.)
+ */
+export function buildProcessPaymentPayload(accepted: Record<string, unknown>): Record<string, unknown> {
+  return {
+    scheme: accepted.scheme,
+    network: accepted.network,
+    amount: accepted.amount,
```
theagenticguy: **Protocol bug — `amount` does not exist on a v1 challenge.** The v1 `accepts` entry names this field `maxAmountRequired` (`specs/x402-specification-v1.md` §5.1.2, "Required"); only v2 renamed it to `amount` (`specs/x402-specification-v2.md` §5.1.2).

Since `extractChallenge` accepts both versions and this builder is shared by both, a v1 challenge yields `amount: undefined` in the `ProcessPayment` payload — the amount is silently dropped rather than erroring. Read `accepted.amount ?? accepted.maxAmountRequired`, or branch on version.

Same root cause one field over: v1 has no top-level `resource` object (it is a *string* on each `accepts` entry), so `challenge.resource` on line 105 falls back to `{}` for v1.

plugins/aws-agent-payments/src/x402.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  const version = String(parsed.x402Version ?? parsed.version ?? "1");
+  const accepts = parsed.accepts as Record<string, unknown>[] | undefined;
+
+  if (!accepts || accepts.length === 0) {
+    throw new Error("x402 challenge has no accepts entries");
+  }
+
+  // Choose the first accepts entry (could be extended to match network preferences)
+  const accepted = accepts[0];
```
theagenticguy: **`accepts[0]` is picked unconditionally, and `networkPreferences` is dead config.** The comment acknowledges the gap ("could be extended to match network preferences"), but the consequence is concrete: `networkPreferences` is declared in `openclaw.plugin.json`'s `configSchema`, threaded through `config.ts:32` and `:60`, written by `setupPaymentInfrastructure`, and then never read by any selection logic. A user who sets it gets no behavior change.

The x402 reference client (`typescript/packages/legacy/x402/src/client/selectPaymentRequirements.ts`) filters `accepts` by supported scheme/network and prioritizes USDC, only falling back to index 0 when nothing matches. Neither spec documents `accepts` as server-preference-ordered, so index 0 is not a safe default.

Concrete failure: a merchant that lists Solana mainnet first and Base Sepolia second gets a payment attempt on Solana from a Base-only wallet, surfacing as the `NetworkMismatch` error `references/debugging.md` documents — and the documented fix ("update `networkPreferences`") cannot work, because nothing reads it.

plugins/aws-agent-payments/src/x402.ts
```diff
@@ -0,0 +1,203 @@
...
+  url: string,
+  headerName: string,
+  headerValue: string
+): Promise<{ status: number; contentType: string; body: string; url: string }> {
+  const response = await fetch(url, {
+    method: "GET",
+    headers: {
+      [headerName]: headerValue,
+    },
+    redirect: "follow",
```
theagenticguy: **Security — the payment proof follows cross-origin redirects.** `redirect: "follow"` combined with a header set on the initial request means Node's `fetch` replays `Payment-Signature`/`X-PAYMENT` to whatever `Location` points at, including a different origin. Verified locally against two servers:

```
[{"srv":"A","hdr":"SECRET_PROOF"},{"srv":"B","hdr":"SECRET_PROOF"}]
```

Server B never issued the 402 and still receives the signed authorization. A malicious or compromised paywall can harvest a valid EIP-3009 authorization by 302-ing the paid request elsewhere.

To be fair on scope: I checked both core specs, both HTTP transport specs, the reference `x402-fetch`/`x402-axios` wrappers, and the May 2026 "Five Attacks on x402" paper (arXiv 2605.11781) — none of them address redirect leakage, so this is not a documented spec requirement. It is still worth handling in a payments plugin. Use `redirect: "manual"` on the replay and re-attach the header only when the redirect target's origin matches the origin that issued the challenge.

`probeUrl` (line 21) has a milder version of the same issue: the challenge may arrive from a different origin than the one requested, and `buildProcessPaymentPayload` never checks `challenge.resource.url` against the requested URL.

plugins/aws-agent-payments/src/payments.ts
```diff
@@ -0,0 +1,351 @@
...
+      }
+    }
+
+    // Get remaining balance from availableLimits
+    let remainingUsd: number | null = null;
+    if (session.availableLimits?.availableSpendAmount?.value) {
+      remainingUsd = parseFloat(session.availableLimits.availableSpendAmount.value);
+    }
+
+    const usable = !expired;
```
theagenticguy: **A drained session reports `usable: true`.** `usable` is derived from expiry alone, so a session whose `availableSpendAmount` has reached zero is still advertised as usable.

`remainingUsd` is computed ten lines up and then not consulted. That directly contradicts the tool description in `src/index.ts:42` ("whether the session is usable ... or whether `create_payment_session` should be called instead") and the SKILL.md Step 1 contract, which routes the agent onward whenever `usable: false`.

Failure path: session has $0.00 left and 90 minutes to live → status returns `usable: true` → agent skips session creation → `get_paid_content` burns a `ProcessPayment` call, fails, and the agent lands in the "Still 402 after payment" branch instead of just minting a new session. Fold the balance in: `usable = !expired && (remainingUsd === null || remainingUsd > 0)`.

plugins/aws-agent-payments/src/payments.ts
```diff
@@ -0,0 +1,351 @@
...
+    paymentManagerArn = createPmResp.paymentManagerArn!;
+    paymentManagerId = createPmResp.paymentManagerId!;
+  }
+
+  // Step 2: Create credential provider
+  // Note: This requires the control-plane CreatePaymentCredentialProvider API
+  // which may need specific permissions
+  const credProvResp = await controlClient.send(
+    new (await import("@aws-sdk/client-bedrock-agentcore-control")).CreatePaymentCredentialProviderCommand({
+      name: `cdp_${Date.now()}`,
```
theagenticguy: **This name violates the constraint your own docs state.** `references/setup.md:74`: "**Credential Provider names**: lowercase alphanumeric + hyphens only (`[a-z0-9-]+`). NO underscores." This generates `cdp_1785270422413` — underscore, and 17 characters of epoch millis. If setup.md is right, every `setup_x402_payments` call fails validation at step 2, after the payment manager has already been created at step 1 (see the next note on partial failure).

Use `cdp-${Date.now()}`. Also worth reconciling `references/debugging.md:56`, which tells users the tool "auto-generates unique names with random suffixes" — `Date.now()` is a timestamp, not random, so two concurrent setups in the same millisecond collide.

plugins/aws-agent-payments/src/payments.ts
```diff
@@ -0,0 +1,351 @@
...
+  const dataClient = new BedrockAgentCoreClient({ region });
+  const { CreatePaymentInstrumentCommand } = await import("@aws-sdk/client-bedrock-agentcore");
+  const instrResp = await dataClient.send(
+    new CreatePaymentInstrumentCommand({
+      paymentManagerArn,
+      paymentConnectorId: connectorId,
+      paymentInstrumentType: "EMBEDDED_CRYPTO_WALLET",
+      paymentInstrumentDetails: {
+        embeddedCryptoWallet: {
+          network: "ETHEREUM",
```
theagenticguy: **Hardcoded `"ETHEREUM"` ignores the `network` parameter.** The `network` argument (default `eip155:84532`, Base Sepolia) reaches only the funding-instructions string on line 335 and the `networkPreferences` echo on line 348. The wallet instrument is always created on `"ETHEREUM"`.

This is the mechanism behind a README claim I could not reconcile: the table advertises Base Mainnet, Solana Devnet, and Solana Mainnet as supported, but no code path can produce a Solana instrument. Either plumb `network` through to `embeddedCryptoWallet` or narrow the README to what setup actually creates.

Two more issues in this function worth addressing together:

- **No partial-failure handling.** Steps 1–4 create four AWS resources with no rollback and no idempotency. A failure at step 3 leaves an orphaned credential provider holding the user's CDP keys, and re-running creates yet another. The `catch` in `src/index.ts:287` swallows this into a generic error message with no record of what was created.
- **`existingPms[0]` (line 265) silently adopts an arbitrary payment manager.** `ListPaymentManagers` is unfiltered, so in an account with several managers this attaches the connector to whichever the API returns first. Filter by name or require the ARN explicitly.

plugins/aws-agent-payments/src/index.ts
```diff
@@ -0,0 +1,301 @@
...
+    name: "setup_x402_payments",
+    description:
+      "Set up x402 payment infrastructure from scratch. Creates the Payment Manager, " +
+      "Credential Provider, Connector, and Wallet Instrument using AWS AgentCore Payments " +
+      "and Coinbase CDP. Requires a Coinbase CDP API key (get from portal.cdp.coinbase.com). " +
+      "Call this when the agent detects x402 payments are not configured. " +
+      "The user must provide their CDP API key ID, secret, and email.",
+    parameters: Type.Object({
+      cdp_api_key_id: Type.String({ description: "Coinbase CDP API Key ID (UUID format)" }),
+      cdp_api_key_secret: Type.String({ description: "Coinbase CDP API Key Secret (base64 Ed25519 key)" }),
```
theagenticguy: **Long-lived wallet credentials as model-visible tool parameters.** `cdp_api_key_secret` and `wallet_secret` are declared as tool inputs, which means the agent host puts them in the model's context and, on most hosts, in a persisted transcript and any request logs. These are not scoped session tokens — the wallet secret is the signing key.

Note the tool description already says the user "must provide their CDP API key ID, secret, and email," so the intended flow is the user pasting a signing key into a chat turn.

Prefer reading these from the environment or a local credentials file at execute time and keeping them out of the tool schema. If they must stay parameters, at minimum mark them sensitive so the host can redact, and document the exposure in `references/setup.md`. This is the highest-value fix in the PR for a plugin whose failure mode is spending someone's money.

Related, at `src/config.ts:87`: `mkdir`/`writeFile` use default permissions for `~/.x402/config.json`. That file holds ARNs, instrument ID, and the live session ID rather than keys, but `mode: 0o700`/`0o600` costs nothing.

plugins/aws-agent-payments/src/index.ts
```diff
@@ -0,0 +1,301 @@
...
+        );
+        headerName = "Payment-Signature";
+      } else {
+        headerValue = paymentResult.headerValue;
+        headerName = "X-PAYMENT";
+      }
+
+      return json({
+        header: { [headerName]: headerValue },
+        valid_seconds: 60,
```
theagenticguy: `valid_seconds: 60` is a hardcoded constant, while the actual validity window comes from the signed authorization's `validBefore` (and the challenge's `maxTimeoutSeconds`, which `buildProcessPaymentPayload` already forwards). SKILL.md turns this number into agent guidance ("Payment headers expire in ~60 seconds"), so a challenge with a shorter or longer window makes that guidance wrong in whichever direction hurts. Derive it: `validBefore - now`.

plugins/aws-agent-payments/skills/x402-payments/references/debugging.md
```diff
@@ -0,0 +1,76 @@
...
+1. Call `get_payment_session_status` to confirm
+2. Ask user to approve a new session
+3. Call `create_payment_session` with desired cap/TTL
+
+### "Still 402 after payment"
+
+**Cause:** The payment was processed but the merchant's on-chain settlement is transient — the signature hasn't propagated yet.
+
+**Fix:**
+- The tool automatically retries. If it persists:
```
theagenticguy: **"The tool automatically retries" is not accurate.** `get_paid_content` makes exactly one `ProcessPayment` call and one replay; on a second 402 it returns an error object immediately (`src/index.ts:136-143`). There is no retry anywhere in `src/`.

This matters more than a doc nit because the surrounding text attributes the failure to settlement propagation delay and tells the reader to wait for a retry that will not happen. Either implement a bounded retry with backoff or rewrite the fix steps around what the tool does.

Two more items in this file to verify while you are here: the `aws bedrock-agentcore-control delete-credential-provider` invocation on line 58 (please confirm the command name and flags against the shipped CLI), and the diagnostic field names on line 66 (`remainingBalance`, `expiresAt`) which do not match what `get_payment_session_status` actually returns (`remaining_usd`, `minutes_left`).

plugins/aws-agent-payments/openclaw.plugin.json
```diff
@@ -0,0 +1,28 @@
+{
```
theagenticguy: **This is the part that needs a maintainer decision before the code review matters.** The PR introduces two new repo-wide patterns — a runtime plugin manifest and a bundled TypeScript implementation — and `CONTRIBUTING.md` is explicit: "For **new plugins** or **major changes** to the repository, contributors must first open an [RFC] before doing any work." I searched all issues and found no RFC for x402, payments, or agent payments, though there is precedent for exactly this (#163 aws-transform, #148 containers, #55 Amazon Location).

Concrete gaps that an RFC would settle:

- **No schema, so no validation.** `schemas/` has manifests for Claude, Codex, marketplace, MCP, and skill frontmatter. `openclaw.plugin.json` matches none, so `mise run lint:manifests` never sees it and nothing keeps the five names in `contracts.tools` in sync with `src/index.ts`.
- **No build path.** `package.json` declares `main: dist/index.js` and `openclaw.extensions: ["./dist/index.js"]`, but has no `scripts.build`, and `dist/` is gitignored at repo root. Nothing in the repo compiles `src/`, and CI has no Node/TypeScript step — `mise run build` is markdown lint, dprint, reference validation, and security scanners. As committed, this plugin cannot produce the artifact it points at. (`npx tsc --noEmit` does pass cleanly against the pinned deps, so the source itself is sound.)
- **No tests.** The repo has no JS test framework (#226 is still open for the Python equivalent), so 976 lines of payment-executing code arrive with zero automated coverage. For code that moves money, that is the gap I would want closed first.
- **Third-party registry in the critical path.** `README.md:48` and the PR body point installs at ClawHub. Please confirm that is settled for an awslabs repo.

None of this is an argument against the capability — x402 auto-pay is a genuinely useful plugin. It is an argument that "runtime plugins" is a repo-architecture change that wants its own RFC and its own CI, separate from shipping this one plugin.

plugins/aws-agent-payments/README.md
```diff
@@ -0,0 +1,162 @@
...
+```bash
+codex plugin marketplace add awslabs/agent-plugins
+```
+
+Then install **aws-agent-payments** from the Plugins panel.
+
+### OpenClaw (Stateful Autonomous Agent)
+
+```bash
+openclaw plugins install clawhub:@aws/aws-agent-payments
```
theagenticguy: This install command does not currently resolve — `npm view @aws/openclaw-x402-payments` and `@aws/aws-agent-payments` both return 404. Worth marking as planned rather than documenting it as available, especially since the PR body describes ClawHub publishing as future work.

Also flagging a naming inconsistency across the four manifests, since a publish pipeline will trip over it: `package.json` says `@aws/openclaw-x402-payments` v**1.0.3**, both plugin manifests say `aws-agent-payments` v**1.0.0**, `openclaw.plugin.json` uses id `aws-agent-payments` with no version, and the config example on line 57 keys the plugin as `x402-payments`. Pick one identity and one version.

plugins/aws-agent-payments/skills/x402-payments/SKILL.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Verification
+
+The skill succeeded if the agent:
+1. Detected the 402
+2. Paid transparently
+3. Returned the paid content to the user
+
+The user should never see "Payment Required" as a final error.
```
theagenticguy, line 183: **The three reference files are unreachable from here.** `mise run validate:refs` flags all of them:

```
ORPHAN  plugins/aws-agent-payments/skills/x402-payments/references/debugging.md
ORPHAN  plugins/aws-agent-payments/skills/x402-payments/references/protocol.md
ORPHAN  plugins/aws-agent-payments/skills/x402-payments/references/setup.md
```

SKILL.md never links them, so the progressive-disclosure model in `docs/DESIGN_GUIDELINES.md` is defeated: the agent has no way to know `references/debugging.md` exists when it hits `NetworkMismatch`. (The task exits 0 on orphans, so this is not a CI failure — it is still a real defect.) Add pointers at the relevant steps, e.g. "See references/debugging.md" in Step 4 and "See references/setup.md" under First-Time Setup.

plugins/aws-agent-payments/.mcp.json
```diff
@@ -0,0 +1,8 @@
+{
+  "mcpServers": {
+    "awsknowledge": {
```
theagenticguy: The `awsknowledge` MCP server is declared but nothing in SKILL.md or the references invokes it. `docs/DESIGN_GUIDELINES.md` says to include "only servers this plugin needs" and exclude "optional nice-to-have integrations." Drop it, or add the skill instructions that use it.

theagenticguy (changes requested):
Thanks for this — x402 auto-pay is a capability the repo does not have today, the v2 flow is coherent end to end, and the source typechecks cleanly against its pinned dependencies (`npx tsc --noEmit` passes). Requesting changes because the build gates fail, because the v1 protocol path looks untested, and because the two new repo patterns need a maintainer decision first.

I ran the repo's own gates on this branch. Four hard failures:

| Gate | Result |
|---|---|
| `mise run lint:manifests` | **FAIL** — `.codex-plugin/plugin.json` missing required `interface` |
| `mise run fmt:check` | **FAIL** — 4 unformatted files (`dprint fmt` fixes) |
| `mise run lint:md` | **FAIL** — non-spec `version` key, description parsed as 2 chars, MD029/MD031/MD032 |
| `mise run lint:cross-refs` | pass, but plugin is absent from **both** marketplace manifests |
| `mise run validate:refs` | 0 broken links; all 3 reference files orphaned |
| `mise run validate:size` | pass — SKILL.md 110 lines, ~1172 tokens, "ideal" |

The one gitleaks finding (`admin:admin` in `.github/workflows/security-scanners.yml`) is pre-existing on main, not from this PR. GitHub only ran the title and contributor-statement checks here, so the Build workflow's failures are not yet visible on the PR.

**Correctness, verified against the specs in `coinbase/x402`:**

1. **v1 payment header is wrong twice over** — sent as raw JSON where `specs/transports-v1/http.md` requires base64, and missing the `{x402Version, scheme, network, payload}` envelope from §5.2. No v1 server accepts it.
2. **v1 amount is silently dropped** — the field is `maxAmountRequired` in v1, and the code reads `accepted.amount` (v2's name) for both versions.
3. **Retry guidance can double-spend** — SKILL.md says `ProcessPayment` is idempotent via `client_token`, but the SDK auto-generates a fresh token per call, so a retry is a second payment.
4. **A drained session reports `usable: true`** — `remainingUsd` is computed and then ignored.
5. **`networkPreferences` is dead config** — declared, plumbed, written, never read; `accepts[0]` is taken unconditionally.
6. **Payment proof follows cross-origin redirects** — reproduced locally; a paywall can 302 a signed authorization to an origin that never issued the challenge.
7. **Wallet signing keys are model-visible tool parameters**, so they land in transcripts and logs.

Given #1 and #2, I would guess the validation against `x402-test.genesisblock.ai` covered only v2. If so, consider dropping v1 from this PR instead of shipping an advertised-but-untested path.

**Process:** `CONTRIBUTING.md` requires an RFC before work begins on a new plugin or a major repo change, and I could not find one for this. Both new patterns here — the `openclaw.plugin.json` runtime manifest and the bundled `src/` implementation — are repo-architecture changes: no schema validates the manifest, no `scripts.build` exists to produce the `dist/index.js` it points at (and `dist/` is gitignored), CI has no Node or TypeScript step, and there is no JS test framework, so 976 lines of money-moving code land with no automated coverage. Splitting this into an RFC for the runtime-plugin pattern plus a skill-only plugin that ships now would de-risk both halves.

Inline comments have the specifics, including exact line numbers and repro steps. Happy to re-review promptly once the gates are green.

plugins/aws-agent-payments/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,28 @@
+{
```
peterjiang-dev, line 1: Do we need claude and codex support for now? If we have not tested against them, anyway we can remove it and only supports OpenClaw?

wirjo:
Thanks for the thorough review @theagenticguy — this is excellent feedback. Pushed a v2 that addresses all points:

## Changes Made

### Structural (following your recommendation to split)

**Removed all runtime code** — `src/`, `openclaw.plugin.json`, `package.json`, `tsconfig.json`, `.mcp.json` are all gone. This is now a **skill-only plugin** matching the existing repo pattern. The runtime plugin pattern (with executable TypeScript) will be proposed via a separate RFC as you suggested.

### Build Gates Fixed

- ✅ `.codex-plugin/plugin.json` now includes full `interface` metadata (displayName, shortDescription, longDescription, defaultPrompt, category, capabilities)
- ✅ Plugin registered in both `.claude-plugin/marketplace.json` and `.agents/plugins/marketplace.json`
- ✅ SKILL.md frontmatter: single-line quoted description (>20 chars), `version` moved under `metadata`
- ✅ `tools/generate_codex_manifests.py` updated with `INTERFACE_METADATA` entry
- ✅ Files formatted with `dprint fmt`

### Correctness Fixes

- ✅ **Dropped v1 claims** — only v2 was validated e2e, so protocol.md and SKILL.md now focus on v2 exclusively
- ✅ **Fixed false idempotency claim** — removed the "safe to retry" guidance. Added note that SDK auto-generates fresh `clientToken` per call, so retries are separate payments unless a stable token is derived from the challenge
- ✅ **Fixed drained session guidance** — SKILL.md Step 1 now explicitly says to check balance, and Step 4 error handling says to verify session balance before retrying
- ✅ **Removed `networkPreferences` dead config** — no longer referenced (runtime code removed)
- ✅ **Redirect security** — protocol.md now documents that replay should use `redirect: "manual"` and verify origin matches
- ✅ **Credential handling** — setup.md and SKILL.md now explicitly state wallet secrets must NEVER be tool parameters. Read from env/config at execution time.

### Reference Files (Orphan Fix)

- ✅ SKILL.md now links all three references:
  - Step 4: "see [references/debugging.md](references/debugging.md)"
  - First-Time Setup: "see [references/setup.md](references/setup.md)"
  - Protocol Details section: "see [references/protocol.md](references/protocol.md)"

### MCP Server

- ✅ Removed `.mcp.json` — `awsknowledge` was declared but unused by the skill

## What's Deferred (to RFC)

- Runtime plugin manifest (`openclaw.plugin.json`)
- TypeScript reference implementation (`src/`)
- Build/test pipeline for executable code
- Schema for runtime manifests

These are repo-architecture changes that warrant their own RFC as you noted. The skill-only plugin ships now and provides value independently.

## Remaining Open

- **RFC:** Will open separately for the runtime-plugin pattern
- **Naming:** Consolidated on `aws-agent-payments` everywhere (plugin id, manifests, marketplace)

Issue #218, opened 2026-07-06
pre-commit: ruff-check fails with real lint errors (F841/E402/F821) in non-template scripts


scottschreckengaust:
Part of #213.

## Problem

`ruff-check` (configured with `--fix --exit-non-zero-on-fix`) reports genuine lint violations in tracked, non-template Python files. Error codes observed: `F841` (assigned-but-unused), `E402` (import not at top), `F821` (undefined name).

## Affected files (non-template)

```
plugins/deploy-on-aws/scripts/lib/post_process_drawio.py
plugins/sagemaker-ai/skills/finetuning/templates/rlvr_reward_function_source_template.py
plugins/sagemaker-ai/skills/model-evaluation/scripts/reward_function_source_template.py
tools/validate-urls.py
```

(Note: the `code_templates/*.py` files also appear in ruff output but are covered separately by #214 — they fail to parse rather than lint.)

## Reproduce

```bash
mise exec -- pre-commit run ruff-check --all-files
```

## Proposed fix

Triage each finding: `F821` (undefined name) may be a real bug worth verifying; `F841`/`E402` are typically safe to fix directly or via `ruff check --fix`. For the `*_source_template.py` files that are code fragments, consider a per-file `# ruff: noqa` or excluding them if they are not standalone modules.

<sub>Filed with assistance from Claude Code while implementing #211.</sub>

Issue #213, opened 2026-07-06
Pre-existing `pre-commit run --all-files` failures (tracking)


scottschreckengaust:
## Summary

`pre-commit run --all-files` currently fails on a clean checkout of `main` (verified at commit `7a17df7`). CI does **not** run the full pre-commit suite — it runs `mise run build` — so these failures are latent and only surface when a contributor runs the hooks locally. This is a tracking issue for the distinct, independent failures; each root cause has its own issue with details and a proposed fix.

## Why this matters

Contributors who install the git hooks (`pre-commit install`) or run `mise run pre-commit` hit failures unrelated to their change, which trains people to bypass hooks with `--no-verify`. Fixing these restores a clean local signal.

## How to reproduce

```bash
git checkout main
mise exec -- pre-commit run --all-files
```

## Per-hook results on clean `main`

| Hook | Result | Root cause |
|------|--------|-----------|
| `check-ast`, `check-builtin-literals`, `debug-statements`, `ruff-format` | Failed | 12 `code_templates/*.py` are intentionally non-runnable and don't parse |
| `check-executables-have-shebangs` | Failed | 6 `.md` files carry the executable bit |
| `check-shebang-scripts-are-executable` | Failed | 2 `.py` scripts have a shebang but no executable bit |
| `check-json` | Failed | `aws4-shapes.json` has a duplicate key (`iot`) |
| `check-vcs-permalinks` | Failed | non-permalink `github.com/.../blob/main/...` URLs in 2 docs |
| `ruff-check` | Failed | real lint errors (F841/E402/F821) in non-template `.py` files |
| `bandit` | Failed (exit 2) | references a missing `.bandit-baseline.json` |
| `no-commit-to-branch` | Failed | **not a bug** — only "fails" because the reproduction runs on `main`; the hook exists to block that. Excluded. |

Sub-issues are linked in the comments below.

<sub>Filed with assistance from Claude Code while implementing #211.</sub>

scottschreckengaust:
Sub-issues filed (one per root cause):

- #214 — 12 `code_templates/*.py` fail Python-parse hooks (`check-ast`, `check-builtin-literals`, `debug-statements`, `ruff-format`)
- #215 — incorrect executable bits (6 `.md` marked +x; 2 `.py` scripts missing +x) → `check-executables-have-shebangs`, `check-shebang-scripts-are-executable`
- #216 — duplicate JSON key `iot` in `aws4-shapes.json` → `check-json`
[...]

PR #244, opened 2026-07-25
fix(deploy-on-aws): merge duplicate iot key in aws4-shapes.json

### LINKED ISSUE #216: pre-commit: check-json fails — duplicate key `iot` in aws4-shapes.json
Part of #213.

## Problem

`check-json` fails to decode a tracked JSON file because of a duplicate key:

```
plugins/deploy-on-aws/scripts/lib/aws4-shapes.json: Failed to json decode (Duplicate key: iot)
```

Duplicate keys are valid to most lenient parsers (last one wins) but are rejected by the strict `check-json` hook and are a latent correctness bug — whichever `iot` block is shadowed is silently ignored.

## Reproduce

```bash
mise exec -- pre-commit run check-json --all-files
```

## Proposed fix

Inspect the two `iot` entries in `aws4-shapes.json`, merge/deduplicate them (keeping the intended definition), and confirm the file still decodes. If the file is generated, fix the generator so it does not emit duplicate keys.

<sub>Filed with assistance from Claude Code while implementing #211.</sub>

TheSaiEaranti:
Fixes #216

## Problem

`plugins/deploy-on-aws/scripts/lib/aws4-shapes.json` defined the `iot` category twice:

| | `fillColor` | shapes |
| --- | --- | --- |
| first definition | `#1A9C37` | 59 |
| second definition | `#3F8624` | 17 |

The two lists overlap by only 14 entries. `check-json` rejects the file outright, and lenient parsers keep the last definition — so the 45 shapes unique to the first definition were silently dropped.

This is not only a lint failure. `validate_drawio.py` flattens every category's `shapes` into `valid_shapes` (lines 25-30) and rejects any shape not in that set (line 225), so valid diagrams were being failed:

```
$ validate_drawio.py iot.drawio
VALIDATION FAILED for iot.drawio:
  ERROR: Invalid AWS4 shape references: iot_bicycle, iot_shadow.
         Check the aws4-shapes.json registry for valid shape names.
```

## Fix

Merge both definitions into a single `iot` entry holding the sorted union — 62 shapes.

- `valid_shapes` goes from 1057 to 1100: **43 shape names become valid again**, and none are removed. (45 were unique to the first definition; 2 of those, `iot_device_gateway` and `iot_over_the_air_update`, were already reachable through another category.)
- The same diagram now passes:

```
$ validate_drawio.py iot.drawio
  INFO: AWS4 shapes used: 2 unique shapes across 1 diagram(s).
VALIDATION PASSED for iot.drawio.
```

### On `fillColor`

I kept `#3F8624`, for three reasons: it is the value that was already in effect, it is shared with five other categories (`storage`, `customer_enablement`, `cloud_financial_management`, `migration_modernization`), and `#1A9C37` appears nowhere else in this registry nor anywhere in the upstream `Sidebar-AWS4.js`. `fillColor` is not consulted by `validate_drawio.py`, so this affects documentation only — happy to switch if `#1A9C37` was the intended IoT colour.

### On the generator

The file header describes it as auto-generated from draw.io's `Sidebar-AWS4.js` and `aws4.xml`, and the issue asks that the generator be fixed if one exists. No generator for this file is committed to the repository — `validate_drawio.py` is the only thing that references it — so the fix is applied to the artifact. If the generator lives elsewhere it will need the same de-duplication, or it will reintroduce this on the next regeneration.

## Verification

- `json.load` with a duplicate-detecting `object_pairs_hook` — no duplicate keys remain (this is what `check-json` enforces).
- `dprint check` on the file — exit 0.
- Shape lists stay sorted, matching 30 of the 31 categories.
- Diff is confined to the `iot` region: 22 insertions, 41 deletions, one file.

Scoped to #216 only; the ruff failures from the same tracking issue (#213) are #218 and are untouched here.

---


TheSaiEaranti:
Not stale, still ready for review. Rebased onto latest main just now and re-validated that aws4-shapes.json parses with no duplicate keys anywhere in the file. Happy to adjust anything if a maintainer wants a different shape for the merged entry.

Issue #214, opened 2026-07-06
pre-commit: 12 `code_templates/*.py` files fail Python-parse hooks (check-ast, ruff-format)


scottschreckengaust:
Part of #213.

## Problem

Twelve `code_templates/*.py` files under `plugins/sagemaker-ai/skills/` are intentionally non-runnable templates (they contain placeholder/templating syntax that is not valid Python). As a result they fail every pre-commit hook that parses Python:

- `check-ast` — Failed
- `check-builtin-literals` — Failed
- `debug-statements` — Failed
- `ruff-format` — Failed (`error: Failed to parse ...`)

## Affected files

```
plugins/sagemaker-ai/skills/dataset-transformation/code_templates/transformation.py
plugins/sagemaker-ai/skills/finetuning/code_templates/dpo.py
plugins/sagemaker-ai/skills/finetuning/code_templates/rlaif_builtin.py
plugins/sagemaker-ai/skills/finetuning/code_templates/rlaif_custom_prompt.py
plugins/sagemaker-ai/skills/finetuning/code_templates/rlvr.py
plugins/sagemaker-ai/skills/finetuning/code_templates/sft.py
plugins/sagemaker-ai/skills/model-deployment/code_templates/deploy-nova-bedrock.py
plugins/sagemaker-ai/skills/model-deployment/code_templates/deploy-nova-sagemaker.py
plugins/sagemaker-ai/skills/model-deployment/code_templates/deploy-oss-bedrock.py
plugins/sagemaker-ai/skills/model-deployment/code_templates/deploy-oss-sagemaker.py
plugins/sagemaker-ai/skills/model-evaluation/code_templates/custom_scorer_evaluator.py
plugins/sagemaker-ai/skills/model-evaluation/code_templates/llmaaj_evaluator.py
```

## Reproduce

```bash
mise exec -- pre-commit run check-ast --all-files
```

## Proposed fix

Exclude `**/code_templates/**` from the Python-parsing hooks (`check-ast`, `check-builtin-literals`, `debug-statements`, `ruff`) in `.pre-commit-config.yaml`, since these files are documentation templates rather than importable modules. Alternatively, rename them to a non-`.py` extension if they are never meant to be executed.

<sub>Filed with assistance from Claude Code while implementing #211.</sub>

PR #245, opened 2026-07-25
fix: resolve ruff-check lint errors in non-template Python files

### LINKED ISSUE #218: pre-commit: ruff-check fails with real lint errors (F841/E402/F821) in non-template scripts

TheSaiEaranti:
Fixes #218

## Scope

`ruff check` (v0.15.0, the pinned pre-commit rev) reports **91** findings on `main`. They split cleanly in two:

| | findings | files | |
| --- | --- | --- | --- |
| `code_templates/*.py` | 46 (44 `invalid-syntax`, 2 `E402`) | 12 | **#214** — untouched here |
| everything else | **45** (`F541` 35, `F841` 5, `F401` 4, `F821` 1) | 12 | **this PR** |

Non-template findings are now **0**. The 46 belonging to #214 are unchanged.

Two notes on how this differs from the issue description, which was written on 2026-07-06:

- The affected non-template file list has grown from 4 to 12 as the repo has gained scripts.
- `E402` is no longer among the non-template codes — both remaining `E402` findings are inside `code_templates/`, so they belong to #214. The non-template codes are now `F541`, `F841`, `F401` and `F821`.

## Mechanical fixes

Applied with `ruff check --fix` (safe fixes only; no `--unsafe-fixes`):

- **`F541` (35)** — dropped the `f` prefix from f-strings with no placeholders.
- **`F401` (2 of 4)** — removed `typing.Union` from `validate_custom_metrics.py` and `INT` from `test_safe_query.py`. I checked that `INT` is not reachable through that file's `globals()`-based fallback test discovery, which selects only callables named `test_*`; `INT_INJECTIONS` is a separate name and is still used.

## Reviewed individually

**`tools/validate-urls.py` — `F821` undefined name `httpx`**
This is a PEP 723 inline script that imports `httpx` lazily inside `check_url` and `main`, so the module-scope annotation `client: "httpx.AsyncClient"` had nothing to resolve against. Added a `TYPE_CHECKING`-guarded import rather than a `noqa`, which keeps the lazy-import behaviour intact — I verified the module still imports with `httpx` absent, and the annotation is preserved.

**`verify-traceability.py` — `F841` `overall_pct`**
`generate_html` computed `overall_total`, `overall_impl` and `overall_pct`, but only `rules_pct` and `reqs_pct` are ever rendered. Removing just `overall_pct` would have left the other two unused, so all three go. This looks like an intended "overall" dashboard stat that was never wired into the HTML — worth a follow-up if you want that row displayed rather than deleted.

**`post_process_drawio.py` — `F841` `actor_x2`**
Removed. Flagging what it hints at, since it may be worth its own issue: the overlap test on the next line is

```python
if actor_x >= cloud_x and actor_x < cloud_x2:
```

which only detects an actor whose **left edge** is inside the AWS Cloud. An actor starting left of the cloud and extending into it is not detected. `actor_x2` is exactly the term a full interval-overlap test would need, so this reads like an unfinished check. I have not changed that behaviour here — it is a layout change, not a lint fix, and should be your call.

## Templates

The three `*_source_template.py` files are customer-facing starter templates (`"INITIAL SUGGESTION ONLY - MUST BE CUSTOMIZED"`), where `from typing import Dict, Any, List, Optional # For type hints` and the example variables are deliberate scaffolding for the reader. Stripping them would remove the thing the template exists to provide, so each gets a file-level directive instead, as the issue suggests:

```python
# Starter template: the imports and example variables below are intentional
# scaffolding for the reader to build on, so unused-name rules do not apply.
# ruff: noqa: F401, F841
```

## Verification

- `ruff check .` → non-template findings 0; remaining 46 are all `code_templates/*` (#214).
- All 12 changed files parse (`ast.parse`) and byte-compile.
- `tools/evals/databases-on-aws/dsql/scripts/test_safe_query.py` → **63 passed, 0 failed**.
- `validate-urls.py` imports at module scope with `httpx` not installed.
- `verify-traceability.py` and `post_process_drawio.py` behave identically to `main` on import (the latter raises the same pre-existing `defusedxml` `AttributeError` in my environment both before and after).

**Out of scope:** `ruff-format` fails repo-wide on `main` — 25 files, including files this PR touches. That count is **25 before and 25 after**, so nothing here adds to it. Running the formatter would turn this into an unrelated 25-file diff, so I left it alone; happy to file it separately if it is not already tracked under #213.

---

### DIFF
--- a/plugins/aws-transform/skills/aws-transform/scripts/verify-traceability.py
+++ b/plugins/aws-transform/skills/aws-transform/scripts/verify-traceability.py
@@ -361,9 +361,6 @@ def generate_html(result: dict) -> str:
 
     rules_pct = _pct(s["implemented_rules"], s["total_rules"])
     reqs_pct = _pct(s["implemented_requirements"], s["total_requirements"])
-    overall_total = s["total_rules"] + s["total_requirements"]
-    overall_impl = s["implemented_rules"] + s["implemented_requirements"]
-    overall_pct = _pct(overall_impl, overall_total)
 
     # Build per-function rows for summary table
     func_rows = ""
@@ -666,7 +663,7 @@ def main():
     # Print summary to stdout
     s = result["summary"]
     print(f"\n{'='*60}")
-    print(f"TRACEABILITY VERIFICATION SUMMARY")
+    print("TRACEABILITY VERIFICATION SUMMARY")
     print(f"{'='*60}")
     print(f"Business Rules (captured):           "
           f"{s['implemented_rules']}/{s['total_rules']} "
@@ -675,7 +672,7 @@ def main():
           f"{s['implemented_requirements']}/{s['total_requirements']} "
           f"({_pct(s['implemented_requirements'], s['total_requirements'])})")
     print(f"{'─'*60}")
-    print(f"Scope: Chapters 1–8 of specification files only")
+    print("Scope: Chapters 1–8 of specification files only")
     print(f"{'='*60}")
 
     missing_rules = s["missing_rules"]
@@ -688,8 +685,8 @@ def main():
             print(f"\n⚠  {missing_reqs} REQ-* identifiers NOT found in chapters "
                   f"1–8 of any specification file.")
     else:
-        print(f"\n✅ All business rules and REQ-* identifiers found in "
-              f"chapters 1–8 of specification files.")
+        print("\n✅ All business rules and REQ-* identifiers found in "
+              "chapters 1–8 of specification files.")
 
     # Write HTML dashboard
     dashboard_html = generate_html(result)
--- a/plugins/deploy-on-aws/scripts/lib/post_process_drawio.py
+++ b/plugins/deploy-on-aws/scripts/lib/post_process_drawio.py
@@ -199,7 +199,6 @@ def fix_placement(tree: ET.ElementTree, verbose: bool = False) -> int:
             cloud_x = aws_cloud_geom["x"]
             cloud_x2 = cloud_x + aws_cloud_geom["w"]
             actor_x = g["x"]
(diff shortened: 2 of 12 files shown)


anwesham-lab (approved):

TheSaiEaranti:
Rebased onto latest main to keep this current after the 1.5.0 sync. Re-ran the pinned ruff 0.15.0 against both this branch and main: main still shows the same 45 non-template findings, this branch shows zero, and the rebase introduced no new ones. Still matches what anwesham-lab approved.

PR #251, opened 2026-08-07
feat(aws-transform): add --mode aws-managed and --resume-incomplete

ntarakad-aws:
## Summary

- Document `--mode aws-managed` for schedule create and remote analysis
- Add aws-managed execution reference
- Rename `--resume` to `--resume-incomplete` in batch execution
- Bump version to 1.6.0


jdonboch (dismissed):

theagenticguy (commented):
Reviewed the full diff plus the surrounding reference files on the branch. This holds together well:

- Version bump is consistent across all three manifests (`marketplace.json`, `.claude-plugin/plugin.json`, `.codex-plugin/plugin.json`).
- Every relative link in the touched files resolves, including the new `continuous-modernization-aws-managed-execution.md`, and the `#1b-create-an-aws-managed-schedule---mode-aws-managed` anchor matches GitHub's slug for that heading.
- The `--resume` → `--resume-incomplete` rename is complete — no stale `--resume` remains on the `remote analysis` surface (the remaining `atx --resume` hits are the unrelated conversation-resume command).
- The aws-managed story is consistent across SKILL.md, analysis, guide, schedule, batch, and the routing file — the "Batch/Fargate is NOT the no-infrastructure option" correction is applied everywhere it previously implied otherwise, and the guide's routing table / step flow / summaries all agree that aws-managed skips Step 3.
- The new reference file's frontmatter (`remote-aws-managed`) matches the `remote-ec2`/`remote-batch` naming convention, and its rejected-flags/error tables agree with the schedule doc's.

Two non-blocking items inline: one question about the removed `--expression`/`--timezone` escape hatch (want to confirm the CLI actually dropped it in 1.6.0, since the skill goes to great lengths to never deny a shipped capability), and one markdown-rendering nit introduced in the error table.

One pre-existing observation, no action needed in this PR: the remote/schedule surfaces use `rapid-techdebt-analysis` while the local `analysis run`/`list` docs use `tech-debt-quick`; that naming split predates this change but might deserve a note somewhere if both are canonical on their respective surfaces.

plugins/aws-transform/skills/aws-transform/references/continuous-modernization-schedule.md
```diff
@@ -198,39 +264,42 @@ atx ct remote status --group <ec2-group-id>
...
-| Customer supplies their own cron                  | `--expression "cron(...)"` (hidden flag; cron only, no `at()`)              |
+| Customer says                                     | Route to                                                                      |
+| ------------------------------------------------- | ----------------------------------------------------------------------------- |
+| "weekly tech-debt scan", "monthly security audit" | `schedule create --recurrence weekly:MONDAY` / `monthly:1`                    |
+| "run it nightly"                                  | `schedule create --recurrence daily`                                          |
+| "remediate these findings on Friday"              | NOT schedulable — run `atx ct remote remediation` when they're ready          |
+| "scan AND auto-fix on a cadence"                  | Schedule the analysis; remediation stays a manual `remote remediation` step   |
+| Customer supplies their own cron                  | Map it to the nearest `--recurrence` (daily / weekly:`<DAY>` / monthly:`<N>`) |
```
theagenticguy, line 277: Question on this row plus the deletion of the "Power-user escape hatch" paragraph (`--expression "cron(...)"` / `--timezone`): was the hidden `--expression` flag actually removed from the CLI in 1.6.0, or is it still accepted for customer-managed schedules?

If the flag still exists, this row now instructs the agent to silently coerce a user-supplied cron to the nearest preset, which changes fire semantics (e.g. `cron(0 3 * * ? *)` becomes `daily`, firing at ~creation-time+2min local rather than 03:00) — and the skill would be denying a capability the CLI ships, which is exactly the failure mode the rest of this PR works hard to prevent. If the flag was removed alongside the `--resume-incomplete` rename, this is correct as written and worth confirming in the PR description.

plugins/aws-transform/skills/aws-transform/references/continuous-modernization-schedule.md
```diff
@@ -198,39 +264,42 @@ atx ct remote status --group <ec2-group-id>
...
-| Token not available                                    | Clone credential not staged          | `atx ct remote credentials --source <src> --token <token> --ack`        |
+| Error                                                                           | Cause                                                 | Fix                                                                                                  |
+| ------------------------------------------------------------------------------- | ----------------------------------------------------- | ---------------------------------------------------------------------------------------------------- |
+| Guidance to `atx ct remote provision`                                           | Compute or scheduler infra missing                    | `atx ct remote provision --mode ec2\|batch --execute --ack`                                          |
+| `--recurrence is required`                                                      | Neither cadence flag given                            | Add `--recurrence daily\|weekly:\<DAY\>\|monthly:\<N\>`                                              |
```
theagenticguy: Rendering nit introduced in this reformat: `\<DAY\>` and `\<N\>` are inside code spans, where backslash escapes don't apply — GitHub will render the backslashes literally (`weekly:\<DAY\>`). Only the pipes need escaping for the table. The Routing table above (line 277) has it right with `` `<DAY>` `` unescaped.

plugins/aws-transform/skills/aws-transform/references/continuous-modernization-schedule.md
ntarakad-aws: `--expression` is still accepted by the CLI (hidden from `--help` via `.hideHelp()`) so existing users with it in scripts won't break. However, it's intentionally not advertised — the skill directs toward `--recurrence` as the supported surface for new users. The flag wasn't removed, just de-promoted.

So the behavior here is: existing `--expression` usage continues to work (backwards-compatible), but the skill won't suggest it to new users.

plugins/aws-transform/skills/aws-transform/references/continuous-modernization-schedule.md
ntarakad-aws: Good catch, will fix.

ntarakad-aws:
Thanks for flagging the `rapid-techdebt-analysis` vs `tech-debt-quick` naming inconsistency — you're right that it's inconsistent across the surfaces. Since it predates this PR, I'll raise a follow-up CR with our team to align the naming. No action in this PR.

zerowsw (approved):

theagenticguy (approved):
LGTM

PR #253, opened 2026-08-14
refactor(sagemaker-ai): remove --command option from hyperpod-issue-report

badmin-aws:
Removes the `--command`/`-cmd` option from the `hyperpod-issue-report` skill.

This is a functional removal of a documented, user-facing capability — not a dead-code cleanup. The flag worked end to end when passed: `generate_collector_script` interpolated each value into the generated bash collector script, and each command's output was written to `command_{NN}_<sanitized_name>.txt` inside `${OUTPUT_DIR}`, tarballed, and uploaded to S3 alongside the built-in diagnostics.

No caller in this repo passed it, but the intended invocation path was agent-mediated rather than hardcoded: `SKILL.md` step 1 instructed the agent to collect "Additional commands to run on nodes" from the user, `SKILL.md` advertised `--command` in the `--help` list, and `references/collection-details.md` documented the resulting `command_NN_*.txt` artifact. So agent-visible behavior changes here, and anyone who previously supplied commands through that prompt will no longer see those files in the bundle.

### Why remove it

The diagnostics the skill actually relies on are built into the generated script — `sinfo`, `nvidia-smi`, `nvidia-bug-report.sh`, `df`, `dmesg`, systemd unit status, Slurm config and logs, the EKS log collector — which covers the cases the removed `--help` examples illustrated (`df -h`, `free -h`).

It also removes an unquoted interpolation. The deleted line built the executed command as `f"{cmd} > ..."` with no quoting, while `shlex.quote` was applied only to the adjacent `echo` display line. That script is uploaded to S3 and executed on every targeted node, so shell metacharacters in a `--command` value were shell metacharacters on the fleet. That is inherent to an arbitrary-command feature rather than a defect in it, but deleting the feature does reduce attack surface, and doing so is preferable to bolting validation onto a capability with no callers. A metacharacter allowlist would have had to reject pipes and redirects to mean anything, and an opt-in override flag would just relocate the same path behind a second flag.

### Changes

- Remove `--command`/`-cmd`, the per-command interpolation loop, and its output-filename sanitization.
- Remove the `commands` parameter from `generate_collector_script`, `execute_collection_on_node`, `execute_with_retry`, and `collect_reports`. Two of these were genuinely dead: `execute_collection_on_node` accepted the parameter and never read it (it builds its SSM invocation from `script_s3_uri`), and `execute_with_retry` only threaded it through. `generate_collector_script` and `collect_reports` did use it.
- Remove the option from `SKILL.md` (both the step-1 input list and the `--help` line) and the `command_NN_*.txt` section from `references/collection-details.md`.
- Bump the `sagemaker-ai` plugin `1.2.1` → `1.3.0` in all three manifests. Minor rather than patch, matching the minor #251 (`9545072`) took for renaming `--resume` → `--resume-incomplete`, since `SKILL.md` made this agent-visible rather than merely internal.

### Known gap

`hyperpod-ssm` covers ad-hoc command execution on a node, but it is not a complete substitute: it returns stdout to the caller one node at a time with no `tar`/S3 bundling, so "capture this extra output *inside* the support bundle" has no direct replacement. Given the built-ins already cover the common diagnostics, the practical gap is small — recording it so the tradeoff is explicit.

### Testing

- Generated collector scripts are byte-identical for both cluster types (Slurm 117 lines, EKS 133 lines); every built-in diagnostic, the tarball step, and the S3 upload are unchanged. Verified by generating both and asserting no `command_` or `Running: ` residue remains.
- `generate_collector_script()` now takes no arguments, so the removed path is unreachable programmatically as well as from the CLI.
- `--command` is now an `argparse` error (exit 2).
- Verified live against a single-node Slurm HyperPod cluster end to end: cluster describe → Slurm detection → node enumeration → script upload → SSM session → on-node execution → tarball → upload → summary. The resulting bundle contained 66 entries (6.8 MB) of real diagnostics — Slurm config and logs, `dmesg`, `syslog`, systemd units, `resource_config.json`, and 28 HyperPod cluster-agent logs.
- `python3 -m py_compile` clean; `bandit` reports no issues; `node tools/validate-cross-refs.cjs` reports 0 errors.

#### Acknowledgment

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -242,7 +242,7 @@
         "machine-learning",
         "generative-ai"
       ],
-      "version": "1.2.1"
+      "version": "1.3.0"
     }
   ]
 }
--- a/plugins/sagemaker-ai/skills/hyperpod-issue-report/SKILL.md
+++ b/plugins/sagemaker-ai/skills/hyperpod-issue-report/SKILL.md
@@ -27,7 +27,6 @@ Collect from the user:
 - **S3 path** for report storage (required, e.g. `s3://bucket/prefix`). If the user doesn't have a bucket, create one (e.g., `s3://hyperpod-diagnostics-<account-id>-<region>`)
 - **Issue description** (optional)
 - **Target scope**: all nodes, specific instance groups, or specific node IDs (optional)
-- **Additional commands** to run on nodes (optional)
 
 ### 2. Verify Environment
 
@@ -60,7 +59,7 @@ uv run scripts/hyperpod_issue_report.py \
   --s3-path s3://<bucket>[/prefix]
 ```
 
-Use `--help` for all options including `--instance-groups`, `--nodes`, `--command`, `--max-workers`, and `--debug`. Note: `--instance-groups` and `--nodes` are mutually exclusive. Node identifiers accept instance IDs (`i-*`), EKS names (`hyperpod-i-*`), or Slurm names (`ip-*`).
+Use `--help` for all options including `--instance-groups`, `--nodes`, `--max-workers`, and `--debug`. Note: `--instance-groups` and `--nodes` are mutually exclusive. Node identifiers accept instance IDs (`i-*`), EKS names (`hyperpod-i-*`), or Slurm names (`ip-*`).
 
 ### 4. Present Results
 
--- a/plugins/sagemaker-ai/skills/hyperpod-issue-report/scripts/hyperpod_issue_report.py
+++ b/plugins/sagemaker-ai/skills/hyperpod-issue-report/scripts/hyperpod_issue_report.py
@@ -313,7 +313,7 @@ class HyperPodIssueReportCollector:
         
         return instance_ids
     
-    def generate_collector_script(self, commands: List[str]) -> str:
+    def generate_collector_script(self) -> str:
         """Generate the bash script that will run on each node.
         Instance group and ID are passed as environment variables.
         Script content varies based on cluster type (EKS vs Slurm)."""
@@ -472,24 +472,6 @@ class HyperPodIssueReportCollector:
                 "",
             ])
         
-        # Add each command to the script
-        for i, cmd in enumerate(commands, 1):
-            # Sanitize command for filename - replace problematic characters
-            safe_name = cmd.replace(' ', '_').replace('/', '_').replace('|', '_').replace('>', '_').replace('<', '_').replace('&', '_').replace(';', '_').replace('(', '_').replace(')', '_').replace('$', '_').replace('`', '_').replace('"', '_').replace("'", '_')[:50]
-            output_file = f"command_{i:02d}_{safe_name}.txt"
-
-            # Use shlex.quote() to safely escape the command for display in echo
-            quoted_cmd = shlex.quote(cmd)
-
-            cmd_line = f"{cmd} > \"${{OUTPUT_DIR}}/{output_file}\" 2>&1 || echo \"Command failed with exit code $?\" >> \"${{OUTPUT_DIR}}/{output_file}\""
-
-            script_lines.extend([
-                f"# Command {i}",
-                f"echo 'Running: '{quoted_cmd}",
-                cmd_line,
-                "",
-            ])
-        
         # Add S3 upload logic with new filename format
         script_lines.extend([
             "# Upload results to S3",
@@ -526,7 +508,7 @@ class HyperPodIssueReportCollector:
             raise ValueError("Cluster ID is required for HyperPod SSM targets")
         return f"sagemaker-cluster:{self.cluster_id}_{instance_group_name}-{instance_id}"
     
-    def execute_collection_on_node(self, node: Dict, commands: List[str], script_s3_uri: str) -> Dict:
+    def execute_collection_on_node(self, node: Dict, script_s3_uri: str) -> Dict:
         """Execute the collection script on a single node via SSM using pexpect."""
         instance_id = node['InstanceId']
         instance_group = node.get('NodeGroup', 'unknown')
@@ -770,10 +752,10 @@ class HyperPodIssueReportCollector:
                 except Exception:  # nosec B110 - best-effort cleanup
                     pass
     
-    def execute_with_retry(self, node: Dict, commands: List[str], script_s3_uri: str, max_retries: int = 3) -> Dict:
+    def execute_with_retry(self, node: Dict, script_s3_uri: str, max_retries: int = 3) -> Dict:
         """Execute collection on a node with exponential backoff on throttling errors."""
         for attempt in range(max_retries):
-            result = self.execute_collection_on_node(node, commands, script_s3_uri)
+            result = self.execute_collection_on_node(node, script_s3_uri)
             
             error_msg = result.get('Error', '')
             if 'ThrottlingException' in error_msg or 'Rate exceeded' in error_msg:
@@ -788,7 +770,7 @@ class HyperPodIssueReportCollector:
         
         return result
 
-    def collect_reports(self, commands: List[str], instance_groups: Optional[List[str]] = None, instance_ids: Optional[List[str]] = None, max_workers: int = 16):
+    def collect_reports(self, instance_groups: Optional[List[str]] = None, instance_ids: Optional[List[str]] = None, max_workers: int = 16):
         """Collect reports from all nodes, specific instance groups, or specific instance IDs.
         
         For Slurm clusters, instance_ids can be either:
@@ -848,12 +830,10 @@ class HyperPodIssueReportCollector:
         elif self.cluster_type == 'slurm':
             print(f"Default collections: nvidia-smi, nvidia-bug-report, sinfo, Slurm services, Slurm config, Slurm logs, system logs")
         
-        if commands:
-            print(f"Additional commands: {', '.join(commands)}")
         print("-" * 60)
-        
+
         # Generate and upload the collector script once
-        script_content = self.generate_collector_script(commands)
+        script_content = self.generate_collector_script()
         script_key = f"{self.report_s3_key}/collector_script.sh"
         
         try:
@@ -874,7 +854,7 @@ class HyperPodIssueReportCollector:
         
         with ThreadPoolExecutor(max_workers=max_workers) as executor:
             future_to_node = {
-                executor.submit(self.execute_with_retry, node, commands, script_s3_uri): node
+                executor.submit(self.execute_with_retry, node, script_s3_uri): node
                 for node in self.nodes
             }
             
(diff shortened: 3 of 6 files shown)


ssuday (dismissed):
LGTM

plugins/sagemaker-ai/skills/hyperpod-issue-report/scripts/hyperpod_issue_report.py
```diff
@@ -472,24 +472,6 @@ def generate_collector_script(self, commands: List[str]) -> str:
                 "",
             ])
         
-        # Add each command to the script
-        for i, cmd in enumerate(commands, 1):
```
scottschreckengaust, line 476: This is the load-bearing correction to the PR description.

The description says the change removes "the resulting dead `commands` parameter on `generate_collector_script`, `execute_collection_on_node`, `execute_with_retry`, and `collect_reports`." That's accurate for `execute_collection_on_node` and `execute_with_retry` — both accepted the parameter and never touched it. But it is **not** accurate for `generate_collector_script`: this loop consumed `commands` to emit a `command_{i:02d}_{safe_name}.txt` artifact into `${OUTPUT_DIR}`, which then got tarballed and uploaded to S3. `collect_reports` also printed the list at old line 851.

So the flag worked end-to-end if you passed it. This is a **functional removal of a user-facing capability**, not a dead-code cleanup. I think removing it is defensible — but please reword the description, because "dead parameter" will send the wrong signal to anyone who bisects to this commit wondering where `command_01_*.txt` went.

Worth adding to the description as a genuine upside: the deleted `cmd_line` interpolated `cmd` into the generated script **unquoted** (`f"{cmd} > ..."`), and that script runs as root on every node. `shlex.quote` was applied only to the `echo` display line, never to the executed command. Removing this shrinks the attack surface — a stronger argument for the change than "it was dead."

plugins/sagemaker-ai/skills/hyperpod-issue-report/SKILL.md
```diff
@@ -27,7 +27,6 @@ Collect from the user:
 - **S3 path** for report storage (required, e.g. `s3://bucket/prefix`). If the user doesn't have a bucket, create one (e.g., `s3://hyperpod-diagnostics-<account-id>-<region>`)
 - **Issue description** (optional)
 - **Target scope**: all nodes, specific instance groups, or specific node IDs (optional)
-- **Additional commands** to run on nodes (optional)
```
scottschreckengaust, line 30: This line is why I'd stop short of calling the flag "unused."

Nothing in the plugin *programmatically* passed `--command` — that part of your reasoning is correct. But this line instructed the agent to actively **collect** "Additional commands to run on nodes" from the user during step 1, and line 63 advertised `--command` in the `--help` list. `references/collection-details.md:66-68` then documented the resulting `command_01_<sanitized_name>.txt` artifact.

So the intended invocation path was agent-mediated rather than hardcoded: the agent asks the user, the user supplies commands, the agent forwards them. The flag had no callers in the repo precisely because the caller was meant to be the model at runtime.

That makes this an agent-visible behavior change, which is the main input to the version-bump question I raised on `marketplace.json`.

.claude-plugin/marketplace.json
```diff
@@ -242,7 +242,7 @@
         "machine-learning",
         "generative-ai"
       ],
-      "version": "1.2.1"
+      "version": "1.2.2"
```
scottschreckengaust: Non-blocking — flagging for your judgment, and deliberately not asking you to change it.

`docs/DESIGN_GUIDELINES.md:485` reserves patch for "backward compatible bug fixes," and `:489-507` asks for a one-minor deprecation window before removing user-facing surface. Read strictly, removing a documented flag would want a minor deprecation notice now and removal at 2.0.0.

I checked whether that policy is actually enforced here before raising it, and it isn't: no plugin in this repo has ever taken a major bump, and there are zero deprecation notices anywhere in the tree. The written policy is aspirational. The closest real precedent is #251 (`9545072`, merged one day before this PR), which renamed `--resume` → `--resume-incomplete` — a breaking CLI rename — and took a **minor** (1.5.0 → 1.6.0).

Against that yardstick `1.3.0` would be more symmetric than `1.2.2`, since `SKILL.md:30` means agent-visible behavior changed and not merely dead code. But patch is inside the variance I see in the history, so I'm not going to hold the PR on it. Your call.

All three manifests are consistent at `1.2.2` and no stray `1.2.1` remains — that part is clean. Note that nothing in CI actually verifies this: `tools/validate-cross-refs.cjs` checks name↔path only, and schema validation checks the semver *pattern* but not cross-file agreement. I'll file that as a separate repo issue rather than something for this PR.

scottschreckengaust (commented):
## Code Review: refactor(sagemaker-ai): remove unused `--command` option from hyperpod-issue-report

### Summary

Well-scoped deletion. The removal itself is clean and complete: all four call sites are consistently updated, `import shlex` is still legitimately used (lines 478-479, 532-534) so no dead import is left behind, `python3 -m py_compile` passes, docs are updated in both places, and all three version manifests agree. All 22 CI checks are green. I found **no correctness defects**.

My comments are about **framing and blast radius**, not code quality. The description characterizes this as removing a *dead parameter*, but the parameter was live in `generate_collector_script` — this is a functional removal of a documented user-facing capability. That's a legitimate call to make; it just deserves to be recorded as such rather than as a no-op cleanup, because it affects how the next person reads the history and how the version was bumped.

### Findings

| # | File | Line | Finding | Severity |
|---|------|------|---------|----------|
| 1 | `scripts/hyperpod_issue_report.py` | 476 (removed) | "dead `commands` parameter on `generate_collector_script`" is inaccurate — it was genuinely used to emit `command_NN_<name>.txt` into the bundle | 🟡 Description accuracy |
| 2 | `SKILL.md` | 30 (removed) | The skill instructed the agent to actively *solicit* "Additional commands" from users, so agent-visible behavior changes — not just an unreachable CLI flag | 🟡 Description accuracy |
| 3 | `marketplace.json` + 2 manifests | 245 | Patch bump (1.2.1 → 1.2.2) for a user-facing capability removal; `#251` took a **minor** for a comparable CLI change | 🔵 Nit / non-blocking |
| 4 | — | — | `hyperpod-ssm` is not a complete substitute: it has no tar/S3 bundling, so "extra commands captured *inside* the support bundle" has no replacement | 🔵 Discussion |

On #3, to be explicit that I'm **not** blocking on it: `docs/DESIGN_GUIDELINES.md:485-507` says breaking changes need a major bump plus a one-minor deprecation window, but that policy has never been exercised in this repo — no plugin has ever taken a major bump, and there are zero deprecation notices anywhere. The nearest precedent, #251 (`9545072`, merged a day before this PR), renamed `--resume` → `--resume-incomplete` and took a minor. So patch is a notch light by that yardstick but well inside observed variance. Your call.

On #4: `hyperpod-ssm/scripts/ssm-exec.sh` has no `s3`/`tar` handling — it returns stdout to the caller one node at a time, subject to SSM output truncation. That covers ad-hoc execution well, but not "capture this extra output into the tarball that goes to AWS Support." Given the built-ins already cover the common cases (`df` at line 373, `nvidia-smi`, systemd, dmesg, Slurm logs — the very things the removed `--help` examples showed), the practical gap is small. Flagging it so the tradeoff is explicit rather than implied.

### What Looks Good

- **The genuinely-dead code claim holds for 2 of 4 methods.** `execute_collection_on_node` and `execute_with_retry` really did just thread the parameter through unused — nice catch.
- **Security posture improves.** The removed code interpolated the raw user string unquoted into a bash script (`cmd_line = f"{cmd} > ..."`) that was uploaded to S3 and executed as root on every node. `shlex.quote` was applied only to the `echo` display line, not to the executed command. Intentional for an arbitrary-command feature, but deleting it is a real attack-surface reduction worth noting in the description as a *benefit*.
- **Complete docs cleanup.** No orphaned references to `--command`, `-cmd`, "Additional commands", or `command_01_` remain anywhere in the repo.
- **No leftover cruft.** Every import still used; no unreachable branches; all three version manifests consistent with no stray `1.2.1`.

### Verdict

**Comment — no changes required, two things worth your judgment.** The code is correct and I'd be happy to see it merge. Before you do, consider (a) editing the PR/commit description so it doesn't call a live, documented feature "unused"/"dead" — that framing will mislead whoever bisects this later, and (b) whether `1.3.0` fits better than `1.2.2` for consistency with #251.

<sub>Two repo-level gaps surfaced during this review that are out of scope here and worth separate issues: the semver/deprecation policy in `DESIGN_GUIDELINES.md:481-507` has never been applied in practice, and `tools/validate-cross-refs.cjs` validates name↔path but not version agreement across the three manifests — nothing in CI would catch a version mismatch.</sub>

badmin-aws:
Thanks — findings 1 and 2 were both correct and I've rewritten the PR description and force-pushed an amended commit message (`f8288d8` → `6d976d0`) so the corrected framing lands in history rather than only in the PR body, since bisect was the concern.

**#1 — "dead parameter" was wrong for `generate_collector_script`.** Agreed. It consumed `commands` to emit `command_{NN}_<sanitized_name>.txt` into `${OUTPUT_DIR}`, which was then tarballed and uploaded, and `collect_reports` printed the list. Both the description and commit message now state plainly that this is a functional removal of a documented user-facing capability, and scope the "genuinely unused" claim to the two methods where it holds — `execute_collection_on_node` (accepted it, never read it, builds its SSM invocation from `script_s3_uri`) and `execute_with_retry` (threaded it through only).

**#2 — "unused" overstated it.** Also agreed. `SKILL.md` step 1 instructed the agent to solicit "Additional commands to run on nodes" from the user, so the intended caller was the model at runtime, not a hardcoded call site. The description now says so explicitly, and notes that anyone who previously supplied commands through that prompt will no longer see those files in the bundle. I've dropped "unused" from the title too.

I also took your suggestion on the security framing: the description now records the unquoted interpolation as a benefit of removal rather than leaving the change to read as a no-op cleanup, while noting it was inherent to an arbitrary-command feature rather than a defect in it.

**#3 — bumped to `1.3.0`.** Your reasoning persuaded me over my own: `SKILL.md` step 1 makes this agent-visible, and #251 (`9545072`) taking a minor for the `--resume` → `--resume-incomplete` rename is the closest real precedent. All three manifests are at `1.3.0` with no stray `1.2.x`.

**#4 — you're right that `hyperpod-ssm` isn't a complete substitute.** `ssm-exec.sh` has no `tar`/S3 handling and returns stdout one node at a time, so "capture this extra output *inside* the support bundle" genuinely has no replacement. I've recorded that as a "Known gap" section in the description and in the commit message rather than letting the `hyperpod-ssm` pointer imply full parity. I think the gap is acceptable given the built-ins cover the diagnostics the removed `--help` examples showed, but it should be explicit rather than implied — thanks for pushing on it.

One extra change in the force-push, unrelated to your findings: the previous revision also deleted an epilog example block (`# With custom prefix and additional commands`) outright. I've restored it as `# With custom S3 prefix` minus the flag, which keeps the PR scoped to just the option removal and keeps this file byte-identical to our internal source so future syncs stay clean.

Agreed both repo-level gaps you spotted are out of scope here — thanks for filing them separately.

scottschreckengaust (approved):
LGTM

ssuday (approved):
LGTM

PR #172, opened 2026-05-20
fix(deploy-on-aws): defusedxml.ElementTree compatibility in fixers

leozhad:
## Summary

Closes #154
Closes #167

The 4 post-processing fixer scripts in `plugins/deploy-on-aws/scripts/lib/` all do `import defusedxml.ElementTree as ET` and then reach for `ET.Element`, `ET.ElementTree`, and `ET.indent`. None of these names are re-exported by `defusedxml.ElementTree`, so the scripts crash at import time on Python 3.10+ (eager annotation evaluation) and on systems with older `defusedxml` (no `indent` re-export).

- **#154** (Element / ElementTree): defusedxml.ElementTree is a thin security wrapper that re-exports the parsing helpers (`parse`, `fromstring`) but not the type aliases. With eager annotations, `def f(cell: ET.Element)` raises:

  ```
  AttributeError: module 'defusedxml.ElementTree' has no attribute 'Element'
  ```

  All 4 files in `plugins/deploy-on-aws/scripts/lib/` are affected (per the comment on #154).

- **#167** (indent): `ET.indent()` is the stdlib pretty-printer added in Python 3.9. Older `defusedxml` releases (still common via system Python and distro packages) never re-exported it, so `ET.indent(tree, ...)` raises:

  ```
  AttributeError: module 'defusedxml.ElementTree' has no attribute 'indent'
  ```

## Approach

Pull the three names directly from the stdlib at import time and attach them to the `ET` module object. defusedxml's secure `parse()` remains the actual XML entry point — none of the symbols added here do any parsing:

- `Element` / `ElementTree` are pure type aliases used in annotations and `dict[str, ET.Element]`-shaped indexes
- `indent()` only mutates the in-memory tree's `text` / `tail` whitespace before serialization; no network or external input is involved

Bandit's B405 ("Using Element to parse untrusted XML data is known to be vulnerable to XML attacks") flags the `from xml.etree.ElementTree import` line. The flagged usage is type-only (no parsing), so each import is annotated with a `# nosec B405` comment explaining that defusedxml remains the parser.

## Verification

- All 4 modules import cleanly under Python 3.12:
  ```bash
  python3 -c "import importlib.util; spec=importlib.util.spec_from_file_location('m','<file>'); m=importlib.util.module_from_spec(spec); spec.loader.exec_module(m); print('OK')"
  ```
[...]

Issue #167, opened 2026-05-15
fix_step_badges.py crashes: defusedxml.ElementTree has no indent attribute on older defusedxml


agorokh:
## Summary

[`plugins/deploy-on-aws/scripts/lib/fix_step_badges.py`](https://github.com/awslabs/agent-plugins/blob/main/plugins/deploy-on-aws/scripts/lib/fix_step_badges.py) calls `defusedxml.ElementTree.indent(...)` but `defusedxml.ElementTree` does not export an `indent()` function on older releases. The stdlib `xml.etree.ElementTree.indent()` was added in Python 3.9, and `defusedxml` did not mirror it until a relatively recent release. On systems with an older `defusedxml` (still commonly installed via system Python, distro packages, or pinned envs), the badge post-processor crashes with:

```
AttributeError: module 'defusedxml.ElementTree' has no attribute 'indent'
```

The offending line is [`fix_step_badges.py:486`](https://github.com/awslabs/agent-plugins/blob/main/plugins/deploy-on-aws/scripts/lib/fix_step_badges.py#L486):

```python
import defusedxml.ElementTree as ET
...
ET.indent(tree, space="  ")
```

## Steps to reproduce

1. Install the `deploy-on-aws` plugin from `awslabs/agent-plugins`.
2. In an environment with an older `defusedxml` (one that doesn't re-export `indent`), generate any architecture diagram that triggers the post-processing pipeline.
3. The badge-overlap fixer step crashes with the `AttributeError` above.

## Workaround

Monkey-patch `defusedxml.ElementTree.indent = xml.etree.ElementTree.indent` before invoking the script.

## Suggested fix

`indent()` is a pure write-side, pretty-print operation — there is no XML-parsing security risk in falling back to the stdlib implementation. Keep `defusedxml` for parsing, use stdlib `indent` for output:

```python
import xml.etree.ElementTree as _ET
import defusedxml.ElementTree as ET
...
# defusedxml lacks indent() on older versions; stdlib indent is safe here
# because it's a write-only pretty-print, not parsing.
try:
    ET.indent(tree, space="  ")
except AttributeError:
    _ET.indent(tree, space="  ")
```

Alternatively, just bump the minimum `defusedxml` version in the plugin's requirements to one that ships `indent()`.

Happy to send a small PR with the one-line patch if useful.

eloyred:
Still present in `deploy-on-aws` v1.3.0.

[...]

Issue #154, opened 2026-05-01
deploy-on-aws: defusedxml.ElementTree does not expose Element/ElementTree types


RafaelPinto:
### Describe the bug

Post-processing scripts use ET.Element/ET.ElementTree type annotations from defusedxml.ElementTree, which doesn't expose those classes. All fixer scripts crash at import time.

### Expected Behavior

Scripts should import and run without errors

### Current Behavior

AttributeError: module 'defusedxml.ElementTree' has no attribute 'Element' on all 4 fixer scripts

### Reproduction Steps

pip install defusedxml>=0.7.1 && python3 -c "import defusedxml.ElementTree as ET; print(ET.Element)"

### Possible Solution

Add from xml.etree.ElementTree import Element, ElementTree and replace ET.Element/ET.ElementTree with direct imports

### Plugin Version

d23cc36 (latest main, 2026-04-30)

### Marketplace Version

N/A — installed from git

### AI Assistant

OpenCode

### AI Assistant Version

1.14.31

### OS

Linux (Ubuntu)

### Region experiencing the issue

N/A (local tooling issue)

### Code modification

No

### Service quota

- [x] I have reviewed the service quotas for this construct

RafaelPinto:
## Affected files
- `plugins/deploy-on-aws/scripts/lib/fix_icon_colors.py`
- `plugins/deploy-on-aws/scripts/lib/fix_nesting.py`
- `plugins/deploy-on-aws/scripts/lib/fix_step_badges.py`
- `plugins/deploy-on-aws/scripts/lib/post_process_drawio.py`

PR #259, opened 2026-08-25
fix(deploy-on-aws): repair defusedxml diagram fixers

shakirjames:
## Summary

- postpone runtime annotation evaluation in all four draw.io fixer modules so `defusedxml.ElementTree` remains the parser without resolving exports it does not provide
- replace the unavailable `ET.indent` call with a small parser-free formatter and make the dynamic loader safe for postponed dataclass annotations
- add regression coverage to the existing `mise run validate` path, with no new dependency

## Problem

With the declared `defusedxml>=0.7.1` dependency, the shipped diagram entrypoints fail at import time because `defusedxml.ElementTree` does not expose `Element`, `ElementTree`, or `indent`. This makes the deploy-on-aws post-processing hook fail before it can validate a generated diagram.

This addresses the runtime failure described in #154 and #167 and supersedes the closed approach in #172 without monkeypatching parser exports or reintroducing an unsafe parser import.

## Verification

- `mise run validate`
- `mise run lint`
- `mise run fmt:check`
- focused unittest suite: 4 passed
- Ruff 0.15.0: clean
- Bandit 1.9.3: clean
- vendored Semgrep scan: 789 rules, 0 findings
- all four shipped entrypoints executed successfully with `defusedxml==0.7.1`
- the shipped hook post-processed and validated a real 2,188 x 904 AWS draw.io diagram end to end

A complete local `mise run build` could not start because the temporary Mise bootstrap rejected its uv release attestation. The exact lint, validation, formatting, focused security, and real-diagram checks above were run directly; canonical GitHub CI remains the package-wide build authority.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -78,6 +78,10 @@ run = [
 # VALIDATION
 # =============
 
+[tasks."test:deploy-on-aws-diagrams"]
+description = "Run deploy-on-aws diagram runtime regression tests"
+run = "uv run --with-requirements plugins/deploy-on-aws/scripts/requirements.txt python -m unittest discover -s plugins/deploy-on-aws/scripts/tests -p 'test_*.py'"
+
 [tasks."validate:refs"]
 description = "Detect broken links and orphaned reference files"
 run = "uv run tools/validate-references.py"
@@ -93,6 +97,7 @@ run = "uv run tools/validate-urls.py"
 [tasks.validate]
 description = "Run all validation checks"
 run = [
+    { task = "test:deploy-on-aws-diagrams" },
     { task = "validate:refs" },
     { task = "validate:size" }
 ]
--- a/plugins/deploy-on-aws/scripts/lib/_xml_format.py
+++ b/plugins/deploy-on-aws/scripts/lib/_xml_format.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python3
+"""Format an already-parsed XML tree without importing an XML parser."""
+
+from __future__ import annotations
+
+
+def indent_tree(tree, space: str = "  ") -> None:
+    """Add deterministic whitespace indentation to an in-memory XML tree."""
+    _indent_element(tree.getroot(), level=0, space=space)
+
+
+def _indent_element(element, *, level: int, space: str) -> None:
+    current_indent = "\n" + (space * level)
+    child_indent = current_indent + space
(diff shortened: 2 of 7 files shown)


eloyred:
This is still broken in the released plugin, and the fix in this PR still applies.

Reproduced on `deploy-on-aws` v1.3.0 with `defusedxml` **0.7.1** — which is both what the plugin's own [`scripts/requirements.txt`](https://github.com/awslabs/agent-plugins/blob/main/plugins/deploy-on-aws/scripts/requirements.txt) asks for (`defusedxml>=0.7.1`) and the latest stable release on PyPI; `0.8.0` exists only as `rc1`/`rc2`. There is no version a user can install that avoids this, so following the documented prerequisite exactly still lands on the bug.

Running the four fixers over a freshly generated diagram:

| script | result |
| --- | --- |
| `fix_icon_colors.py` | `AttributeError: module 'defusedxml.ElementTree' has no attribute 'indent'` |
| `fix_step_badges.py` | same, whenever it has badges to move |
| `fix_nesting.py` | exits 0 — nothing to write for this file |
| `post_process_drawio.py` | exits 0 — same |

### Why this has been easy to miss

The crash only happens **when a fixer actually has something to write**. Against an already-clean diagram all four exit 0, so a smoke test passes.

And `fix_step_badges.py` prints its success line before raising:

```console
$ python3 scripts/lib/fix_step_badges.py diagram.drawio
Badges moved: 2
Traceback (most recent call last):
  File ".../fix_step_badges.py", line 494, in <module>
    main()
  File ".../fix_step_badges.py", line 486, in main
    ET.indent(tree, space="  ")
AttributeError: module 'defusedxml.ElementTree' has no attribute 'indent'
```

The repositioning is computed, reported as done, and then discarded — `tree.write()` is never reached. Anyone not reading the traceback sees `Badges moved: 2` and keeps an unfixed diagram.

### Reopening

@krokoko @laithalsaadoon — mentioning you per the stale bot's own instructions for reopening, as the two reviewers requested on both this PR and #172.

This PR, #172 before it, and issues #154 and #167 have all been closed by the stale bot without a human review, while the bug remains in the released plugin. The approach here looks right to me — parser-free formatting, no new dependency, regression coverage on the existing `mise run validate` path — but I am happy to help get it over the line if it needs changes.

PR #261, opened 2026-08-27
feat(databases-on-aws): add native DSQL foreign key support

davidrz15:
## Summary

- Add native Aurora DSQL foreign-key guidance using referenced and referencing terminology.
- Update DDL, migration, ORM, and troubleshooting guidance to preserve supported foreign keys and remove the application-layer referential-integrity workflow.
- Add focused foreign-key eval coverage and update plugin metadata.
- Keep the functional-eval harness unchanged.

Part of the foreign-key rollout with:
- https://github.com/awslabs/aurora-dsql-orms/pull/598
- https://github.com/awslabs/aurora-dsql-tools/pull/147

## Validation

- `mise run build`
- Python tests: 65 passed
- JSON, Markdown, formatting, manifest, cross-reference, reference, and size validation
- Bandit, Gitleaks, Checkov, and Grype

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/agent-plugins/blob/main/LICENSE).

PR #262, opened 2026-08-28
fix(databases-on-aws): harden DSQL functional eval harness

davidrz15:
## Summary

- migrate the DSQL functional corpora to a versioned schema with explicit regex or LLM grading
- fail closed around subject isolation, tool/result correlation, transact blocking, redaction, and artifact promotion
- close review gaps around bounded deterministic scans, process identity, output ownership/locking, descriptor-relative staging, and unsafe grader false positives
- preserve trusted artifact/recovery metadata while bounding adversarial escaped-key and nested-secret redaction
- add focused regression coverage and document the supported runner workflows

This is split from #261 so the eval harness hardening can be reviewed independently from the foreign key skill content.

## Testing

- `mise run test:python` (`125 passed`)
- `mise run fmt:check`
- `mise run build` (lint, formatting, tests, Bandit, Semgrep, Gitleaks, Checkov, and Grype)
- complete DSQL authoring/security self-review using Code Review and PR Review Toolkit lenses
- `git diff --check`

## Review follow-ups resolved

- bounded adversarial DDL, sensitive-key, escaped-key, and mapping-collision scans
- correlated the real `transact.sql_list` payload with lint source/fixed SQL and rejected comment-based bypasses
- rejected contradictory legacy safety claims plus literal f-string, `%`, and `+` SQL interpolation
- disabled print-mode prompt suggestions and rejected nested-CLI option injection through model arguments
- hardened process cleanup, output-directory and lock inode identity, descriptor-relative staging, and promotion recovery
- preserved trusted artifact schema/control-plane fields while redacting untrusted values and deeply escaped credentials
- replaced the interpolation detector's catastrophic backtracking pattern with bounded lookaheads
- restored escaped-separator redaction and preserved evidence after unterminated escaped assignments
- made output-lock and staging cleanup idempotent under injected unlock/removal failures
- restricted trusted artifact-value bypasses to scalar constants

## Known limitation

macOS does not expose a supported primitive that reliably tracks a rapid double-fork plus `setsid()` escape. The runner improves `kqueue`/`libproc` error handling and cleanup, but does not claim that unsupported case is fully contained.

### DIFF
--- a/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
+++ b/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
@@ -2123,6 +2123,12 @@ SENSITIVE_KEY = re.compile(
     + r"|endpoint|hostname|account|arn)",
     re.IGNORECASE,
 )
+ESCAPED_SENSITIVE_KEY_TOKEN = re.compile(
+    r"(?i)(?<![A-Za-z0-9])(?:"
+    r"authorization|cookie|credential|passphrase|password|"
+    r"private[_-]?key|secret|session|signature|hmac|token|api[_-]?key"
+    r")(?![A-Za-z0-9])"
+)
 SAFE_TELEMETRY_KEYS = {
     "cache_creation_input_tokens",
     "cache_read_input_tokens",
@@ -2485,7 +2491,7 @@ def _redact_nested_escaped_sensitive_values(value: str) -> str:
     cursor = 0
     search_from = 0
     length = len(value)
-    while match := SENSITIVE_KEY.search(value, search_from):
+    while match := ESCAPED_SENSITIVE_KEY_TOKEN.search(value, search_from):
         search_from = match.end()
         if match.start() < cursor:
             continue
--- a/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
+++ b/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
@@ -17,6 +17,7 @@ import os
 import re
 import select
 import selectors
+import shlex
 import signal
 import shutil
 import stat
@@ -137,16 +138,97 @@ MAX_JSON_NESTING = 100
 MAX_JSON_NUMBER_LENGTH = 100
 MAX_JSON_KEY_LENGTH = 500
 MAX_REDACTION_INPUT = 2 * MAX_ARTIFACT_TEXT
+MAX_TIMEOUT_SECONDS = 24 * 60 * 60
 DEFAULT_JUDGE_TIMEOUT_SECONDS = 60
 REDACTION_KEY = os.urandom(32)
 EXPLICIT_ENVIRONMENT_SECRETS: set[str] = set()
 ENVIRONMENT_NAME = re.compile(r"[A-Za-z_][A-Za-z0-9_]*")
+UNSAFE_PASSTHROUGH_ENVIRONMENT = re.compile(
+    r"^(?:BASH_ENV|ENV|GCONV_PATH|IFS|NODE_OPTIONS|PERL5OPT|PERLLIB|"
+    r"PYTHONHOME|PYTHONINSPECT|PYTHONPATH|PYTHONSTARTUP|RUBYOPT|"
+    r"LD_.+|DYLD_.+)$"
+)
 UNSAFE_PLUGIN_PATH = re.compile(r"[,()\r\n]")
 INFORMATIONAL_EVENT_TYPES = {
+    "prompt_suggestion",
     "rate_limit_event",
     "tool_progress",
     "tool_use_summary",
 }
+TRUSTED_ARTIFACT_KEYS = frozenset({
+    "PreToolUse",
+    "artifact_type",
+    "assertions",
+    "assertion_failures",
+    "command",
+    "count",
+    "description",
+    "duration_seconds",
+    "eval_id",
+    "eval_name",
+    "expected_output",
+    "evidence",
+    "expectations",
+    "failed",
+    "focus",
+    "graded_total",
+    "grading",
+    "grader",
+    "grading_protocol_version",
+    "hooks",
+    "id",
+    "infrastructure_error",
+    "infrastructure_errors",
+    "input",
+    "is_error",
+    "judge_cost_usd",
+    "judge_duration_seconds",
+    "judge_errors",
+    "matcher",
+    "messages",
+    "name",
+    "overall_pass_rate",
+    "pass_rate",
+    "passed",
+    "prompt",
+    "reason",
+    "requested_total",
+    "result_text",
+    "results",
+    "returncode",
+    "run_configuration",
+    "schema_version",
+    "skill_name",
+    "status",
+    "stderr",
+    "summary",
+    "text",
+    "timing",
+    "tool_calls",
+    "tool_results",
+    "tool_use_id",
+    "total",
+    "total_cost_usd",
+    "total_duration_seconds",
+    "total_evals",
+    "total_expectations",
+    "total_failed",
+    "total_passed",
+    "truncated",
+    "truncated_failures",
+    "truncations",
+    "turn_count",
+    "type",
+    "usage",
+    "version",
+})
+TRUSTED_ARTIFACT_VALUE_KEYS = frozenset({
+    "artifact_type",
+    "grader",
+    "grading_protocol_version",
+    "schema_version",
+    "status",
+})
 
 SUBPROCESS_ENV_KEYS = {
     "ANTHROPIC_API_KEY",
@@ -200,7 +282,7 @@ class OutputDirectoryLease:
         self.lock_descriptor = lock_descriptor
 
     def assert_identity(self) -> None:
-        """Reject replacement of the claimed output directory."""
+        """Reject replacement of the claimed output directory or lock file."""
         if self.directory_descriptor < 0:
             raise OSError("output directory lease is closed")
         try:
@@ -219,13 +301,41 @@ class OutputDirectoryLease:
             raise OSError(
                 f"claimed output directory was replaced: {self.path}"
             )
+        try:
+            lock_path_stat = os.stat(
+                OUTPUT_LOCK,
+                dir_fd=self.directory_descriptor,
+                follow_symlinks=False,
+            )
+        except OSError as error:
+            raise OSError(
+                f"claimed output lock was replaced: {self.path / OUTPUT_LOCK}"
+            ) from error
+        lock_descriptor_stat = os.fstat(self.lock_descriptor)
+        if (
+            not stat.S_ISREG(lock_path_stat.st_mode)
+            or (lock_path_stat.st_dev, lock_path_stat.st_ino)
+            != (
+                lock_descriptor_stat.st_dev,
+                lock_descriptor_stat.st_ino,
+            )
+        ):
+            raise OSError(
+                f"claimed output lock was replaced: {self.path / OUTPUT_LOCK}"
+            )
 
     def close(self) -> None:
         if self.lock_descriptor >= 0:
-            _release_output_lock(self.lock_descriptor)
+            _release_output_lock(
+                self.lock_descriptor,
+                self.directory_descriptor,
+            )
             self.lock_descriptor = -1
         if self.directory_descriptor >= 0:
-            os.close(self.directory_descriptor)
+            try:
+                fcntl.flock(self.directory_descriptor, fcntl.LOCK_UN)
+            finally:
+                os.close(self.directory_descriptor)
             self.directory_descriptor = -1
 
     def __del__(self):
@@ -233,6 +343,79 @@ class OutputDirectoryLease:
             self.close()
         except OSError:
             pass
+
+
+class DescriptorTemporaryDirectory:
+    """Own a temporary child of an already opened directory descriptor."""
+
+    def __init__(self, parent_descriptor: int, *, prefix: str):
+        self.parent_descriptor = parent_descriptor
+        self.directory_descriptor = -1
+        self.entry_name = ""
+        for _ in range(100):
+            entry_name = prefix + os.urandom(12).hex()
+            try:
+                os.mkdir(
+                    entry_name,
+                    mode=0o700,
+                    dir_fd=parent_descriptor,
+                )
+            except FileExistsError:
+                continue
+            self.entry_name = entry_name
+            break
+        if not self.entry_name:
+            raise FileExistsError(
+                "could not allocate a unique descriptor-relative directory"
+            )
+        try:
+            self.directory_descriptor = _open_directory_at(
+                parent_descriptor,
+                self.entry_name,
+            )
+        except BaseException:
+            self.cleanup()
+            raise
+
+    @property
+    def name(self) -> str:
+        """Return a path to the opened temporary directory's current inode."""
+        if self.directory_descriptor < 0:
+            raise OSError("descriptor-relative temporary directory is closed")
+        proc_path = Path(
+            "/proc/self/fd",
+            str(self.directory_descriptor),
+        )
+        if proc_path.is_dir():
+            return str(proc_path)
+        if hasattr(fcntl, "F_GETPATH"):
+            raw_path = fcntl.fcntl(
+                self.directory_descriptor,
+                fcntl.F_GETPATH,
+                bytes(1024),
+            )
+            current_path = raw_path.split(b"\0", 1)[0]
+            if current_path:
+                return os.fsdecode(current_path)
+        raise OSError(
+            "platform does not expose descriptor-relative directory paths"
+        )
+
+    def cleanup(self) -> None:
+        if not self.entry_name:
+            return
+        try:
+            if self.directory_descriptor >= 0:
+                os.close(self.directory_descriptor)
+                self.directory_descriptor = -1
+            _remove_output_entry(
+                self.parent_descriptor,
+                self.entry_name,
+            )
+        except FileNotFoundError:
+            pass
+        finally:
+            self.entry_name = ""
 
 
 class RedactingArgumentParser(argparse.ArgumentParser):
@@ -437,9 +620,8 @@ class _ProcessTreeMonitor:
                 | select.KQ_EV_ENABLE
                 | select.KQ_EV_CLEAR
             ),
-            # Darwin has exposed NOTE_TRACK constants since 10.5 but the
-            # kernel rejects them with ENOTSUP. Poll libproc around fork
-            # notifications instead.
+            # Darwin still declares NOTE_TRACK, but modern kernels reject it
+            # with ENOTSUP. Poll libproc around fork notifications instead.
             fflags=select.KQ_NOTE_EXIT | select.KQ_NOTE_FORK,
         )
         self.kqueue.control([event], 0, 0)
@@ -809,7 +991,7 @@ def _run_captured(
         raise SystemExit(128 + received_signal)
 
     if threading.current_thread() is threading.main_thread():
-        for signal_number in (signal.SIGTERM, signal.SIGHUP):
+        for signal_number in (signal.SIGTERM, signal.SIGHUP, signal.SIGQUIT):
             previous_signal_handlers[signal_number] = signal.getsignal(signal_number)
             signal.signal(signal_number, terminate_on_signal)
 
@@ -826,7 +1008,12 @@ def _run_captured(
         ):
             cleanup_mask = signal.pthread_sigmask(
                 signal.SIG_BLOCK,
-                {signal.SIGINT, signal.SIGTERM, signal.SIGHUP},
+                {
+                    signal.SIGINT,
+                    signal.SIGTERM,
+                    signal.SIGHUP,
+                    signal.SIGQUIT,
+                },
             )
         try:
             if process is not None:
@@ -1134,6 +1321,7 @@ def _write_transact_guard_plugin(plugin_dir: Path) -> str:
             "description": "Blocks DSQL writes while retaining attempted tool calls",
             "version": "1.0.0",
         },
+        redact=False,
     )
     _write_private_json(
         hooks_dir / "hooks.json",
@@ -1144,12 +1332,14 @@ def _write_transact_guard_plugin(plugin_dir: Path) -> str:
                     "hooks": [{
                         "type": "command",
                         "command": (
-                            "${CLAUDE_PLUGIN_ROOT}/block-transact.py"
+                            f"{shlex.quote(sys.executable)} "
+                            '"${CLAUDE_PLUGIN_ROOT}/block-transact.py"'
                         ),
                     }],
                 }],
             },
         },
+        redact=False,
     )
     script = plugin_dir / "block-transact.py"
     descriptor = os.open(
@@ -1159,7 +1349,7 @@ def _write_transact_guard_plugin(plugin_dir: Path) -> str:
     )
     with os.fdopen(descriptor, "w") as script_file:
         script_file.write(
-            f"#!{sys.executable}\n"
+            "#!/usr/bin/env python3\n"
             "import sys\n"
             f"sys.stderr.write({denial_marker!r} + '\\n')\n"
             "sys.exit(2)\n"
@@ -1188,6 +1378,7 @@ def run_prompt(
         "claude", "-p",
         "--output-format", "stream-json",
         "--verbose",
+        "--prompt-suggestions", "false",
         "--plugin-dir", str(resolved_plugin_dir),
         "--max-turns", str(max_turns),
         "--mcp-config", str(resolved_mcp_config),
@@ -1728,6 +1919,8 @@ def _llm_judge(
         system_prompt,
         "--output-format",
         "json",
+        "--prompt-suggestions",
+        "false",
         "--max-turns",
         "1",
         "--tools",
@@ -1803,6 +1996,22 @@ def _llm_judge(
                     "LLM judge total_cost_usd must be a nonnegative finite number",
                     duration_seconds=duration,
                 )
+        outer_errors = outer.get("errors", [])
+        if (
+            not isinstance(outer_errors, list)
+            or any(not isinstance(item, str) for item in outer_errors)
+        ):
+            return _judge_error(
+                "LLM judge outer errors field must be an array of strings",
+                duration_seconds=duration,
+                cost_usd=judge_cost,
+            )
+        if outer_errors:
+            return _judge_error(
+                "LLM judge reported top-level errors",
+                duration_seconds=duration,
+                cost_usd=judge_cost,
+            )
         if (
             outer.get("subtype") != "success"
             or type(outer.get("is_error")) is not bool
@@ -1948,11 +2157,6 @@ GENERIC_ASSIGNMENT = re.compile(
     r"(?P<prefix>\b(?P<name>[A-Za-z][A-Za-z0-9_-]{2,80})\s*[:=]\s*)"
     r"(?P<value>\"(?:\\.|[^\"\\])*\"|'(?:\\.|[^'\\])*'|[^\s,;]+)"
 )
-ESCAPED_SENSITIVE_VALUE = re.compile(
-    r"(?is)(?:\\+[\"'])?\b"
-    + SENSITIVE_NAME_PATTERN
-    + r"\b(?:\\+[\"'])?\s*\\*[:=]\s*\\+[\"'].*?\\+[\"']"
-)
 BEARER_TOKEN = re.compile(r"(?i)\bbearer\s+[A-Za-z0-9._~+/=-]+")
 AUTHORIZATION_HEADER = re.compile(
     r"(?im)\b(?P<name>authorization|proxy-authorization)\s*:[^\r\n]*"
@@ -2009,10 +2213,6 @@ GENERIC_HOME_PATH = re.compile(
     r"[A-Z]:\\Users\\[^\\\s]+|"
     r"(?<![A-Za-z0-9_])~(?=[/\\])"
     r")"
-)
-JSON_KEY_VALUE = re.compile(
-    r'(?P<prefix>"(?P<key>(?:\\u[0-9a-fA-F]{4}|\\.|[^"\\])*)"\s*:\s*)'
-    r'(?P<value>"(?:\\.|[^"\\])*"|[^\s,}\]]+)',
 )
 DOLLAR_QUOTE_START = re.compile(r"\$[A-Za-z_][A-Za-z0-9_]*\$|\$\$")
 SINGLE_QUOTE_START = re.compile(r"(?:[bBeEnNxX]|[uU]&)?'")
@@ -2249,33 +2449,89 @@ def _redact_private_keys(value: str) -> str:
     return "".join(output)
 
 
-def _redact_unicode_escaped_json_secrets(value: str) -> str:
-    """Redact JSON values whose sensitive key uses Unicode escapes."""
-
-    def replace(match: re.Match) -> str:
-        try:
-            decoded_key = _json_loads(f'"{match.group("key")}"')
-        except (json.JSONDecodeError, JsonValidationError):
-            return match.group(0)
-        if isinstance(decoded_key, str) and _is_sensitive_key(decoded_key):
-            return match.group("prefix") + '"<redacted>"'
-        return match.group(0)
-
-    return JSON_KEY_VALUE.sub(replace, value)
-
-
 def _decode_ascii_unicode_escapes(value: str) -> str:
     """Expose escaped ASCII key names to the credential redactors."""
+    output = []
+    index = 0
+    length = len(value)
+    while index < length:
+        if value[index] != "\\":
+            output.append(value[index])
+            index += 1
+            continue
+        run_start = index
+        while index < length and value[index] == "\\":
+            index += 1
+        if (
+            index + 5 <= length
+            and value[index] == "u"
+            and all(
+                character in "0123456789abcdefABCDEF"
+                for character in value[index + 1:index + 5]
+            )
+        ):
+            codepoint = int(value[index + 1:index + 5], 16)
+            if codepoint < 128:
+                output.append(chr(codepoint))
+                index += 5
+                continue
+        output.append(value[run_start:index])
+    return "".join(output)
 
-    def replace(match: re.Match) -> str:
-        codepoint = int(match.group("codepoint"), 16)
-        return chr(codepoint) if codepoint < 128 else match.group(0)
 
-    return re.sub(
-        r"(?:\\)+u(?P<codepoint>[0-9a-fA-F]{4})",
-        replace,
-        value,
-    )
+def _redact_nested_escaped_sensitive_values(value: str) -> str:
+    """Redact arbitrarily escaped JSON-like credentials in one forward scan."""
+    output = []
+    cursor = 0
+    search_from = 0
+    length = len(value)
+    while match := SENSITIVE_KEY.search(value, search_from):
+        search_from = match.end()
+        if match.start() < cursor:
+            continue
+        index = match.end()
+
+        slash_start = index
+        while index < length and value[index] == "\\":
+            index += 1
+        if index == slash_start or index >= length or value[index] not in "\"'":
+            continue
+        index += 1
+        while index < length and value[index].isspace():
+            index += 1
+        if index >= length or value[index] not in ":=":
+            continue
+        index += 1
+        while index < length and value[index].isspace():
+            index += 1
+
+        slash_start = index
+        while index < length and value[index] == "\\":
+            index += 1
+        if index == slash_start or index >= length or value[index] not in "\"'":
+            continue
+        quote = value[index]
+        index += 1
+
+        closing_end = length
+        while index < length:
+            quote_index = value.find(quote, index)
+            if quote_index < 0:
+                break
+            slash_index = quote_index
+            while slash_index > index and value[slash_index - 1] == "\\":
+                slash_index -= 1
+            if slash_index < quote_index:
+                closing_end = quote_index + 1
+                break
+            index = quote_index + 1
+
+        output.append(value[cursor:match.start()])
+        output.append("<redacted-secret>")
+        cursor = closing_end
+        search_from = max(search_from, cursor)
+    output.append(value[cursor:])
+    return "".join(output)
 
 
 def _environment_secret_values() -> set[str]:
@@ -2321,7 +2577,6 @@ def _redact_text(
     redacted = TERMINAL_CONTROL.sub("", redacted)
     redacted = _redact_private_keys(redacted)
     redacted = _decode_ascii_unicode_escapes(redacted)
-    redacted = _redact_unicode_escaped_json_secrets(redacted)
     for home_path in sorted(
         {
             str(Path.home()),
@@ -2358,7 +2613,7 @@ def _redact_text(
         lambda match: f"{match.group('name')}: <redacted>",
         redacted,
     )
-    redacted = ESCAPED_SENSITIVE_VALUE.sub("<redacted-secret>", redacted)
+    redacted = _redact_nested_escaped_sensitive_values(redacted)
     redacted = AWS_ENV_CREDENTIAL.sub("<redacted-aws-credential>", redacted)
     redacted = SENSITIVE_VALUE.sub("<redacted-secret>", redacted)
     redacted = GENERIC_ASSIGNMENT.sub(
@@ -2382,17 +2637,56 @@ def _redact_text(
     return _truncate_text_head_tail(redacted, MAX_REDACTION_INPUT)
 
 
-def _redact_artifact_value(value, key: str = ""):
+def _redact_artifact_value(
+    value,
+    key: str = "",
+    *,
+    trusted_keys: frozenset[str] = frozenset(),
+):
     """Apply final sink redaction to every string in persisted output."""
+    if key in TRUSTED_ARTIFACT_VALUE_KEYS:
+        return value
     if isinstance(key, str) and _is_sensitive_key(key):
         return "<redacted>"
     if isinstance(value, dict):
-        return _redact_mapping(
-            value,
-            lambda item, item_key: _redact_artifact_value(item, item_key),
-        )
+        redacted = {}
+        next_suffix = {}
+        for item_key, item in value.items():
+            output_key = (
+                item_key
+                if item_key in trusted_keys
+                else _redact_mapping_key(item_key)
+            )
+            suffix = next_suffix.get(output_key, 2)
+            base_key = output_key
+            while output_key in redacted:
+                collision = f"<collision:{suffix}>"
+                output_key = (
+                    _truncate_text(
+                        base_key,
+                        MAX_JSON_KEY_LENGTH - len(collision),
+                    )
+                    + collision
+                    if isinstance(base_key, str)
+                    else f"{base_key}{collision}"
+                )
+                suffix += 1
+            next_suffix[base_key] = suffix
+            redacted[output_key] = _redact_artifact_value(
+                item,
+                item_key,
+                trusted_keys=trusted_keys,
+            )
+        return redacted
     if isinstance(value, (list, tuple)):
-        return [_redact_artifact_value(item, key) for item in value]
+        return [
+            _redact_artifact_value(
+                item,
+                key,
+                trusted_keys=trusted_keys,
+            )
+            for item in value
+        ]
     if isinstance(value, str):
         return _redact_text(value, redact_sql_literals=True)
     return value
@@ -2490,10 +2784,11 @@ def _redact_mapping_key(key):
 def _redact_mapping(value: dict, transform) -> dict:
     """Redact keys and preserve colliding entries under deterministic suffixes."""
     redacted = {}
+    next_suffix = {}
     for item_key, item in value.items():
         base_key = _redact_mapping_key(item_key)
         output_key = base_key
-        suffix = 2
+        suffix = next_suffix.get(base_key, 2)
         while output_key in redacted:
             collision = f"<collision:{suffix}>"
             output_key = (
@@ -2506,6 +2801,7 @@ def _redact_mapping(value: dict, transform) -> dict:
                 else f"{base_key}{collision}"
             )
             suffix += 1
+        next_suffix[base_key] = suffix
         redacted[output_key] = transform(item, item_key)
     return redacted
 
@@ -2678,6 +2974,7 @@ def _message_timeline(
     run_result: dict,
     *,
     omit_result_content: bool = False,
+    text_limit: int = 1000,
 ) -> list[dict]:
     """Preserve redacted assistant/tool event order for temporal assertions."""
     events = []
@@ -2704,7 +3001,7 @@ def _message_timeline(
                             str(block.get("text", "")),
                             redact_sql_literals=True,
                         ),
-                        1000,
+                        text_limit,
                     ),
                 })
             elif block_type == "tool_use":
@@ -3063,10 +3360,7 @@ def _has_positive_match(
     upper_bound: bool = False,
 ) -> bool:
     """Return whether at least one occurrence supports the asserted claim."""
-    matches = list(re.finditer(pattern, value, re.IGNORECASE))
-    if not matches:
-        return False
-    for match in matches:
+    for match in re.finditer(pattern, value, re.IGNORECASE):
         prefix = value[max(0, match.start() - 80):match.start()]
         suffix = value[match.end():match.end() + 80]
         if upper_bound and POSITIVE_UPPER_BOUND.search(prefix):
@@ -3212,19 +3506,29 @@ def _legacy_keyword_match(expectation: str, evidence: str) -> tuple[bool, int, i
         criteria.append((keyword, _match_is_positive(expectation, match)))
 
     matched = 0
+    contradicted = False
     for keyword, expected_positive in criteria:
-        evidence_matches = re.finditer(
+        evidence_matches = list(re.finditer(
             rf"\b{re.escape(keyword)}\b",
             evidence,
             re.IGNORECASE,
-        )
+        ))
         if any(
             _match_is_positive(evidence, match) == expected_positive
             for match in evidence_matches
         ):
             matched += 1
+        if any(
+            _match_is_positive(evidence, match) != expected_positive
+            for match in evidence_matches
+        ):
+            contradicted = True
     return (
-        bool(criteria and matched / len(criteria) >= 0.6),
+        bool(
+            criteria
+            and not contradicted
+            and matched / len(criteria) >= 0.6
+        ),
         matched,
         len(criteria),
     )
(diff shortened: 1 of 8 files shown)


anwesham-lab (dismissed):
largely lgtm, leaving a comment for follow ups we should make, but they can be picked up after merged and I'd like to get the FK support in ASAP as priority

anwesham-lab:
Reviewed this as the harness half of the #261 split. **The split is clean and the blocking findings from #261 are genuinely fixed** — no FK content leaked in (0 FK-related additions across the corpora diffs), so this reviews independently as intended. Recommending merge, with one small fix worth taking here and two follow-ups.

Verified fixed, at head `8bf388cf5f8b0764a457ba02a2bce1214b332980`:

| #261 | Fix in this PR |
|---|---|
| `--bare` stripped the `Skill` tool, so the suite graded the bare model | Flag removed entirely, plus [`_dsql_skill_loaded()`](https://github.com/awslabs/agent-plugins/blob/8bf388cf5f8b0764a457ba02a2bce1214b332980/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py#L1070-L1090) correlating the Skill `tool_use_id` to a non-error result |
| `.claude/.mcp.json` didn't exist; two documented commands exited 1 | Defaults to the shipped config; where the path appears it's documented as gitignored / operator-supplied |
| Hardcoded `--max-turns 10`; exhaustion misfiled as infrastructure | CLI flag, and `error_max_turns` reclassified as `truncated` with an explicit warning |
| Missing/misspelled `llm_judge` silently fell back to keyword grading | Required `grader` key through a [`Grader`](https://github.com/awslabs/agent-plugins/blob/8bf388cf5f8b0764a457ba02a2bce1214b332980/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py#L34-L37) enum that raises on anything else |
| Missing `expectations` graded nothing and exited 0 | `EvalSchemaError` on empty/missing, plus `skill_name` validation |
| `overall_pass_rate` computed only over graded assertions | `requested_total` / `graded_total` split with `truncated_failures` |
| Unresolved `--eval-ids` warned and exited 0 | Now raises `eval IDs not found: [...]` |

The provenance block (`corpus_sha256`, `plugin_tree_sha256`, `selected_eval_ids`) and blocking `mcp__aurora-dsql__transact` behind a generated guard plugin are good additions beyond what was asked for.

---

### Smallest follow-up, could be in this PR: redaction corrupts UUIDs

In [`_redact_text`](https://github.com/awslabs/agent-plugins/blob/8bf388cf5f8b0764a457ba02a2bce1214b332980/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py#L2334-L2344) the 12-digit account rule fires *inside* UUIDs before [`UUID_VALUE`](https://github.com/awslabs/agent-plugins/blob/8bf388cf5f8b0764a457ba02a2bce1214b332980/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py#L1945-L1948) (L2339) can match, destroying the shape; [`PAYMENT_CARD_VALUE`](https://github.com/awslabs/agent-plugins/blob/8bf388cf5f8b0764a457ba02a2bce1214b332980/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py#L1950)'s `\d[ -]*?` then eats the hyphenated leading groups. Executed against this branch:

```
00000000-0000-0000-0000-000000000000  ->  <redacted-payment-card>-0000-<redacted-account>
550e8400-e29b-41d4-a716-446655440000  ->  550e8400-e29b-41d4-a716-<redacted-account>
550e8400-e29b-41d4-a716-4466554400ab  ->  <redacted-uuid>          # only this one is correct
```

Only UUIDs with a hex letter in the final group redact cleanly — including the repo's own all-zeros example convention, which comes out as garbage. Since the DSQL evals are built on example UUIDs, artifacts and judge evidence carry mangled placeholders.

Scope check before anyone worries: this is an **evidence-quality** bug, not a grading bug. I confirmed DDL structure survives redaction intact — column types and constraint clauses pass through — so no eval passes or fails wrongly because of it. Fix is to run `UUID_VALUE.sub` ahead of the account and payment-card rules, or anchor those two to reject matches adjacent to `-`-delimited hex groups.

Two things I suspected and **disproved** by testing, so they need no action: `PAYMENT_CARD_VALUE` is not a ReDoS risk (0.000s at 36 digits), and redaction does not damage the DDL that assertions actually grade.

### Follow-up 1: decompose the test suite

`test_run_functional_evals.py` is 3,237 lines and collects **4 tests** — no classes, no `parametrize`. That's down from 19 focused tests in #261, which is why `mise run test:python` reports 69 rather than main's 84.

To be clear about severity, since "4 tests" reads worse than it is: I measured coverage, and those 4 tests exercise **83% of 2,832 statements** in the runner. The fixes above are genuinely protected against regression. The cost is diagnosability — a failure reports one of four ~800-line tests like `test_grading_correlates_tools_redacts_evidence_and_fails_closed`, and the first failed assert hides everything downstream. Real maintenance cost, but refactoring work on a suite that demonstrably covers the code, so it doesn't need to gate this merge.

### Follow-up 2: `REDACTION_KEY` stability

[`REDACTION_KEY = os.urandom(32)`](https://github.com/awslabs/agent-plugins/blob/8bf388cf5f8b0764a457ba02a2bce1214b332980/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py#L139) is regenerated per process, so SQL-literal placeholders differ between runs and artifacts aren't diffable run-to-run. Deriving it from `corpus_sha256` would keep the pseudonymization property while making artifacts comparable — which is most of the value of writing them.

### Non-blocking, but cheaper to answer now than later

605 → 6,335 lines in the runner. Most of it earns its place; [`_ProcessTreeMonitor`](https://github.com/awslabs/agent-plugins/blob/8bf388cf5f8b0764a457ba02a2bce1214b332980/tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py#L329-L651) is the part I'd question — 322 lines of `kqueue` + `libproc` via `ctypes` + Linux `prctl(37)` subreaper + `/proc` scraping to track descendants escaping the process group, when the subject is `claude -p` launched by a maintainer on their own machine rather than hostile code. The PR body also concedes it doesn't contain the double-fork + `setsid()` case it's built for, so it's carrying permanent complexity for a guarantee it explicitly doesn't provide. Same question for the hand-rolled JSON limits and the SSN / payment-card / JWT / cookie classes, which don't appear in DSQL eval traffic.

Raising it now only because it's the one item that gets *harder* after merge — nobody volunteers to delete working code, and everyone who touches this file next pays the comprehension tax. Entirely a maintainer judgment call, not a merge condition. The `transact` block and credential redaction should stay either way.

---

Note on scope: this was a targeted pass — I verified the #261 findings, measured coverage, and executed the redaction paths, rather than the full sub-agent roster #261 got. That was deliberate since this is just tooling/CI, a test suite, and one consumer.

anwesham-lab (dismissed):

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,6328 @@
...
+        and not bool(result.get("is_error", False))
+    }
+
+    expectations = []
+    grader = Grader(eval_item["grader"])
+    judge_duration_seconds = 0.0
+    judge_cost_usd = 0.0
+    judge_evidence = ""
+    if grader is Grader.LLM_JUDGE:
+        if len(text) > MAX_REDACTION_INPUT:
```
laithalsaadoon, line 4499: Blocking: this `MAX_REDACTION_INPUT` guard is nested inside the `LLM_JUDGE` branch, so the regex grader only bounds `text` by `MAX_CAPTURE_BYTES` (2 MB, line 4004) before handing it to the deterministic rules — three of which are quadratic (see 3552). Hoisting this check above line 4052 gates both graders.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,6328 @@
...
+def _create_table_bodies(value: str) -> list[str]:
+    """Extract active balanced CREATE TABLE bodies, failing on malformed DDL."""
+    masked = _mask_sql_comments(value)
+    bodies = []
+    malformed = False
+    for start_match in CREATE_TABLE_START.finditer(masked):
+        if _sql_example_is_negated(masked, start_match):
+            continue
+        open_index = start_match.end() - 1
+        depth = 0
```
laithalsaadoon, line 3993: Blocking: this balanced-paren walk runs from every `CREATE TABLE (` match and breaks only on `depth == 0`, so unbalanced input never breaks and each of N matches scans to end-of-string.

Measured through `_every_created_table_has_tenant_id` on `"create table t("` repeated: 20 KB = 1.66 s, 50 KB = 10.14 s, 100 KB = 43.51 s. At 30 KB the cost is this function (5.12 s), not the regex (`_mask_sql_comments` = 0.01 s).

Reachable from the shipped corpus: `evals.json` eval 2 is `"grader": "regex"` with "Includes tenant_id column in all tables" → `AssertionRule.TENANT_ID` (4498). Grading has no timeout, so an oversized answer reads as a CI hang. `_has_positive_drop_warning` (3636, 5.53 s at 184 KB) and `_sql_example_is_negated` (3510, 2.88 s at 56 KB) share the shape.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,6328 @@
...
+
+def _redact_unicode_escaped_json_secrets(value: str) -> str:
+    """Redact JSON values whose sensitive key uses Unicode escapes."""
+
+    def replace(match: re.Match) -> str:
+        try:
+            decoded_key = _json_loads(f'"{match.group("key")}"')
+        except (json.JSONDecodeError, JsonValidationError):
+            return match.group(0)
+        if isinstance(decoded_key, str) and _is_sensitive_key(decoded_key):
```
laithalsaadoon: Blocking: `SENSITIVE_KEY` opens with `(?:[A-Za-z0-9]+[_-])*` before a literal alternation, which backtracks quadratically on a long alnum run with no sensitive tail. The other three call sites slice `key[:MAX_JSON_KEY_LENGTH]` (2350, 2484, 2576); this one does not, and its key comes from the `JSON_KEY_VALUE` regex rather than `_json_loads`, so the 500-char bound never applies.

Measured through `_redact_text` on `'"' + "a"*n + '":1'`: 5 KB = 0.24 s, 10 KB = 0.88 s, 30 KB = 8.14 s. `_redact_text` runs per timeline event, per tool call, and per tool result, so this multiplies. Same slice as its siblings should do it.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,6328 @@
...
-    expectations = []
-    use_llm_judge = bool(eval_item.get("llm_judge", False))
+def _is_sensitive_key(value: str) -> bool:
+    normalized = re.sub(r"[^a-z0-9]", "", value.casefold())
+    return (
+        value.casefold() not in SAFE_TELEMETRY_KEYS
+        and (
+            SENSITIVE_KEY.search(value) is not None
+            or any(
+                token in normalized
```
laithalsaadoon: Blocking: `signature`, `sig`, and `hmac` are missing here and from `SENSITIVE_NAME_PATTERN`. Verified on this branch: `_is_sensitive_key` returns False for `X-Amz-Signature`, `signature`, `sig`; True for `X-Amz-Credential`, `X-Amz-Security-Token`, `password`.

`_redact_artifact_value({"X-Amz-Signature": ..., "password": ...})` redacts `password` and passes the signature through verbatim, and the free-text form `X-Amz-Signature=<64 hex>` survives `_redact_text` unchanged. DSQL auth is a SigV4 presigned `DbConnect` URL, so this is the highest-value secret the harness handles, and it lands in `grading.json`/`eval_metadata.json` and on the judge's stdin. `_redact_tool_result_value` catches it via its allowlist; the artifact and judge paths are denylist-only.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,6328 @@
...
-                evidence = "No mention of 8 columns per index limit found"
+        _recover_abandoned_promotions(output_dir)
+        for child in output_dir.iterdir():
+            if not re.fullmatch(r"\.run-[A-Za-z0-9_.-]+", child.name):
+                continue
+            _remove_output_path(child)
+        sibling_run_pattern = re.compile(
+            rf"\.{re.escape(output_dir.name)}\.run-[A-Za-z0-9_.-]+"
+        )
+        for sibling in output_dir.parent.iterdir():
```
laithalsaadoon: Blocking: this iterates the *parent* of the output directory and `rmtree`s anything matching `\.<outdir-name>\.run-*`. The ownership marker and flock cover `output_dir` only, so nothing confirms a matched sibling belongs to this runner.

With `--output-dir /tmp/results`, a pre-existing `/tmp/.results.run-anything` is deleted; if it belongs to another user the delete raises instead and the runner cannot start at all. Writing the ownership marker inside each staged dir and requiring it here would close both.

laithalsaadoon (dismissed):
Reviewed with a focus on path traversal and OS/network-level risk in the new harness. Two things I could not fault: no corpus field reaches a filesystem path (`eval_id` is type-checked as a non-negative int at 3801-3808 and is the only corpus value used as a path component), and no corpus content reaches argv (prompt and judge payload both go over stdin at 1187/1721, `shell=` is never used). The promotion path's `O_NOFOLLOW` + `dir_fd` + flock + inode identity discipline looks right.

Four blocking findings inline. Each performance and redaction claim was measured by importing the module and calling the real function on this branch, not inferred from pattern shape; timings are in the comments.

**Please double-check these before I re-review** — I would value a second opinion on two points specifically:

1. The `_create_table_bodies` timings (comment on line 3552) are from adversarial input (`"create table t("` repeated). Do you consider a 2 MB unbalanced-DDL answer reachable in practice for a `regex` grader eval, or is the realistic ceiling much lower than the constant permits?
2. Whether `signature`/`sig` genuinely belong in `_is_sensitive_key` (line 2033) given `_redact_tool_result_value`'s allowlist already covers the tool-result path. My concern is the artifact and judge paths, which are denylist-only — worth confirming that reading is right.

Non-blocking notes I did not comment on, happy to file separately if useful: env secrets under 12 chars are skipped inside mapping keys (2445); a bare 40-char AWS secret access key has no free-text detector (`AWS_ENV_CREDENTIAL` at 1928 requires a `NAME=` prefix); `--model`/`--judge-model` accept a leading `-` and become CLI flags; `os.killpg` at 659/697 runs after the child is already reaped at 951-957, so it can signal a recycled PID, and `signal_known`/`wait_known` drop the `(pid, start_time)` identity the Linux tracker already collects at 542-549; `expanduser()` at 1246 raises `RuntimeError` (not `OSError`) on an unknown `~user` prefix, which aborts the whole run.

The known-limitation note about macOS double-fork plus `setsid()` is the right call to document rather than paper over.

laithalsaadoon (commented):
Suggested edits for three of the five blocking findings, each applied locally against `0d501ce` and verified before posting. The harness suite passes with all three in place (`python3 -m pytest -q test_run_functional_evals.py`, 4 passed).

Two of my blocking comments I am **withdrawing as written**, because your own test suite documents intent that my suggestions would have broken. I tried both and they failed:

1. **Line 4053 (hoisting the `MAX_REDACTION_INPUT` guard out of the `LLM_JUDGE` branch)** — this breaks `test_grading_correlates_tools_redacts_evidence_and_fails_closed:2044`, which feeds a `2 * MAX_ARTIFACT_TEXT` answer with a forbidden `CREATE INDEX` buried in the middle and requires the regex grader to *fail* it. Grading long transcripts deterministically is deliberate, so the guard is correctly judge-only. The quadratic scanners are still the real problem; the line 3552 suggestion below fixes them at the source instead, which is the better place anyway.

2. **Line 5358 (gating the sibling sweep on the ownership marker)** — this breaks `test_main_covers_both_graders_artifacts_and_incomplete_runs:2689`, which creates an *unmarked* `.results.run-injected` sibling and requires cleanup to remove it. Marker-gating cannot work here: a staging dir abandoned mid-crash may never have been marked. I still think deleting unmarked directories in an unowned parent is a sharp edge, but the fix has to preserve crash cleanup — perhaps confining staging to a subdirectory of the leased output dir, or matching on a per-run random suffix the runner records. Your call on the shape; I withdraw my specific suggestion.

The other two blocking findings (`os.killpg` on a reaped PID, and `expanduser` raising `RuntimeError`) I have left as prose rather than suggestions, since both fixes touch control flow where you will have better context than I do.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,6328 @@
...
+        _table_has_tenant_column(body)
+        for body in bodies
+    )
+
+
+def _has_ordered_positive_stages(value: str, *patterns: str) -> bool:
+    """Require a positive occurrence of every stage in the declared order."""
+    cursor = 0
+    for pattern in patterns:
+        positive_match = None
```
laithalsaadoon, line 4039: Suggested fix for the quadratic body scan: a total scan budget rather than a per-match cap (a per-match cap alone only halves it, since the match count still grows with input).

```suggestion
    for start_match in CREATE_TABLE_START.finditer(masked):
        if _sql_example_is_negated(masked, start_match):
            continue
        open_index = start_match.end() - 1
        depth = 0
        quote = ""
        index = open_index
        if scan_budget <= 0:
            malformed = True
            break
        scan_limit = min(len(masked), open_index + scan_budget)
        while index < scan_limit:
            character = masked[index]
            if quote:
                if character == quote:
                    if index + 1 < len(masked) and masked[index + 1] == quote:
                        index += 2
                        continue
                    quote = ""
                elif character == "\\" and quote == "'":
                    index += 2
                    continue
            elif character in {"'", '"'}:
                quote = character
            elif character == "(":
                depth += 1
            elif character == ")":
                depth -= 1
                if depth == 0:
                    bodies.append(masked[open_index + 1:index])
                    break
            index += 1
        else:
            malformed = True
        scan_budget -= index - open_index
    return [] if malformed else bodies
```

Requires `scan_budget = MAX_SQL_BODY_SCAN` initialized above the loop (line 3548) and `MAX_SQL_BODY_SCAN = 2 * MAX_CORPUS_TEXT` beside the other bounds. Measured on the adversarial input from my earlier comment: **43.51 s -> 0.030 s** at 100 KB. Verified `_every_created_table_has_tenant_id` still returns True for a real multi-table schema with `tenant_id` and False when any table omits it, and `mise run test:python` stays green (4 passed).

Budget exhaustion falls into the existing `malformed` branch, so grading fails closed rather than silently passing an unscanned tail — please confirm that is the semantic you want.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,6328 @@
...
+    """Redact JSON values whose sensitive key uses Unicode escapes."""
+
+    def replace(match: re.Match) -> str:
+        try:
+            decoded_key = _json_loads(f'"{match.group("key")}"')
+        except (json.JSONDecodeError, JsonValidationError):
+            return match.group(0)
+        if isinstance(decoded_key, str) and _is_sensitive_key(decoded_key):
+            return match.group("prefix") + '"<redacted>"'
+        return match.group(0)
```
laithalsaadoon: Suggested fix for the unbounded key reaching the quadratic `SENSITIVE_KEY` scan — same slice the other three call sites already use (2350, 2484, 2576):

```suggestion
        if isinstance(decoded_key, str) and _is_sensitive_key(
            decoded_key[:MAX_JSON_KEY_LENGTH]
        ):
            return match.group("prefix") + '"<redacted>"'
        return match.group(0)
```

Measured through `_redact_text` on a 30 KB alphanumeric key: **8.14 s -> 0.020 s**. Suite still green.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,6328 @@
...
-        elif ch == "}":
-            depth -= 1
-            if depth == 0:
-                return s[start:i + 1]
-    return None
 
+SENSITIVE_NAME_PATTERN = (
+    r"(?:[A-Za-z0-9]+[_-])*(?:authorization|cookie|credential|passphrase|"
+    r"password|private[_-]?key|secret|session|token|api[_-]?key)"
+)
```
laithalsaadoon, line 2123: Suggested fix for the SigV4 gap, adding the long names to the shared pattern:

```suggestion
    r"(?:[A-Za-z0-9]+[_-])*(?:authorization|cookie|credential|passphrase|"
    r"password|private[_-]?key|secret|session|signature|hmac|token|"
    r"api[_-]?key)"
)
```

This also covers the free-text path via `SENSITIVE_VALUE`, which shares the pattern: `X-Amz-Signature=<64 hex>` now redacts instead of passing through. `sig` needs separate handling because a substring rule would swallow `signal`/`design`/`assignment` — an exact-match set works:

```python
SENSITIVE_EXACT_KEYS = {
    "hmac",
    "sig",
}
```

added beside `SAFE_TELEMETRY_KEYS`, with `or normalized in SENSITIVE_EXACT_KEYS` in the `_is_sensitive_key` return and `"signature"` in the token tuple at 2033.

Verified after patching: `X-Amz-Signature`, `signature`, `sig`, `hmac` all True; `signal`, `significant`, `sigma`, `design`, `assignment`, `input_tokens`, `content`, `code`, `status` all still False; `_redact_artifact_value({"X-Amz-Signature": ...})` now yields `<redacted>`. Suite green.

anwesham-lab:
I'll pick up blocking comments separately, to strengthen evals and run evals locally and add to report in a comment in review of FKs to simultaneously unblock that workstream.

anwesham-lab:
Implemented the requested review follow-ups in `fe72bb8`.

Resolved:

- bounded adversarial `CREATE TABLE` and sensitive-key scans;
- added SigV4 `signature` / `sig` / `hmac` redaction without losing benign oversized-key evidence;
- skipped process-group signals after the root is reaped and validated Linux PID start-time identity in fallback cleanup;
- handled `Path.expanduser()` `RuntimeError` as a fail-closed Read-scope violation;
- moved staging under the owned output directory, removed the parent sibling sweep, and made marker/recovery cleanup relative to the leased directory descriptor;
- rejected explicit f-string injection and combined-DDL transaction guidance in deterministic grading.

The two withdrawn suggestions remain unchanged: the complete semantic-redaction limit is still judge-only, and unmarked crash staging is not marker-gated.

Verification:

- `mise run test:python` — 81 passed
- `mise run build` — passed
- direct Bandit scan of the modified runner/tests — no findings
- complete 20-area DSQL review and security roster — completed
- `git diff --check` — clean

Working on running complete automated fleets across various judges on security and other areas.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
anwesham-lab: Preserved as requested after your withdrawal: the complete semantic-redaction limit remains scoped to the LLM judge. The quadratic deterministic scan was fixed at its source in fe72bb8.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
anwesham-lab: Fixed in fe72bb8 with a total CREATE TABLE scan budget and a focused adversarial regression. The strengthened SQL-context scan regression completes within the bound.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
anwesham-lab: Fixed in fe72bb8. Sensitive-key classification is now linear for arbitrary key lengths, preserves benign oversized-key evidence, and redacts sensitive suffixes across text, artifact, judge, and tool-result sinks. Regression coverage includes both behavior and scan time.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
anwesham-lab: Fixed in fe72bb8. X-Amz-Signature, signature, sig, and hmac are redacted across free-text and structured sinks, with false-positive checks for signal/significant/sigma/design/assignment and safe telemetry keys.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
anwesham-lab: Fixed with a different shape in fe72bb8. Run staging now lives inside the owned output directory, parent-sibling sweeping was removed, and marker/recovery cleanup uses the leased directory descriptor. Unrelated .results.run-* siblings are preserved while abandoned internal .run-* staging is still cleaned.

anwesham-lab:
Final DSQL authoring/security convergence review completed at `4476945`.

The contributor branch was rebased onto newer `main` after review. The DSQL
files at rebased head `4476945` are byte-identical to reviewed local head
`fd3af00`; only commit IDs changed.

| # | Confidence | Area | Finding | Disposition | Commit / SHA |
|---|---:|---|---|---|---|
| 1 | 99 | grader-security | Real `transact` payloads use `sql_list`; string-only correlation could miss unsafe execution | Fixed with per-statement fail-closed correlation and comment canonicalization | `82f175a` |
| 2 | 99 | grader-security | Contradictory legacy claims and literal f-string / `%` / `+` interpolation could pass safety assertions | Fixed with contradiction checks and structural interpolation detection | `82f175a` |
| 3 | 99 | output-safety | Lock replacement, path replacement, and reserved-prefix recovery could permit concurrency bugs or delete/strand data | Fixed with directory+file inode locking, descriptor-relative staging, validated recovery state, and recoverable internal preparation names | `82f175a` |
| 4 | 99 | redaction | Trusted schema/control-plane fields could be corrupted; nested or adversarial escaped values could leak or consume superlinear time | Fixed with trusted structural fields, linear escaped-key scans, deep nested-secret handling, and bounded collision logic | `82f175a`, `4476945` |
| 5 | 98 | process/CLI | Prompt-suggestion events, nested CLI option values, startup-control environment variables, and SIGQUIT cleanup were not fully fail-closed | Fixed with explicit CLI flags/validation and cleanup coverage | `82f175a` |
| 6 | 99 | docs/comments | Recovery and advisory-lock documentation overstated sibling cleanup and arbitrary-writer exclusion | Corrected | `82f175a` |

Verification on the reviewed SHA:

- `mise run test:python` — **116 passed**
- `mise run build` — **passed**
- Bandit — no findings
- Gitleaks — no leaks
- Checkov — 71 passed, 0 failed
- Grype — no vulnerabilities
- complete 20-area Code Review / PR Review Toolkit roster plus final convergence pass
- `git diff --check` — clean

No unresolved finding at the review procedure's confidence threshold remains. The documented macOS rapid double-fork plus `setsid()` limitation remains intentionally scoped as previously agreed.

anwesham-lab (dismissed):

anwesham-lab:
## Complete 20+ agent review-fleet report

**Reviewed head:** `4476945`  
**Result:** every merge-relevant finding at the review procedure's confidence
threshold is fixed, explicitly withdrawn by the reviewer, or documented as an
accepted platform limitation. All eight GitHub review threads are resolved.

The contributor branch was rebased after the review commits were pushed. The
DSQL eval files at `4476945` were compared byte-for-byte with reviewed local
head `fd3af00` and are identical; only commit IDs changed.

### Fleet coverage

| # | Review lens | Coverage and outcome |
|---:|---|---|
| 1 | Repository instructions | Verified worktree isolation, `mise` workflow, dependency boundaries, required full build, and no unrelated changes. Complete. |
| 2 | Shallow correctness | Found and fixed real `transact.sql_list` correlation, unrelated-SQL lint acceptance, literal `%` / `+` interpolation, and fail-open malformed payloads. |
| 3 | History / blame | Checked the intent behind process containment, long-answer regex grading, crash recovery, and the documented macOS limitation. No stale semantic regression remains. |
| 4 | Prior PR comments | Revalidated every `theagenticguy` inline and review-body item against the final head. All active requests are fixed; withdrawn suggestions remain intentionally unapplied. |
| 5 | Comment analyzer | Corrected stale advisory-lock, snapshot, sibling-staging, and Darwin `NOTE_TRACK` wording. |
| 6 | Silent-failure hunter | Fixed ignored judge errors, ambiguous/malformed transact input, comment-based lint bypasses, promotion recovery failures, and failure-reporting gaps. |
| 7 | Code simplification | Removed or avoided duplicate assertion/recovery behavior where it could hide production paths; no blocking simplification finding remains. |
| 8 | PR test analyzer | Added focused tests for every validated boundary. Runner coverage is now 51 focused tests; repository total is 116 tests. |
| 9 | Type-design analyzer | Hardened artifact schemas, trusted control-plane fields, CLI argument validation, promotion state, and invalid-state rejection. |
| 10 | General correctness reviewer | Confirmed the real MCP transact shape, lint-result ordering, per-statement correlation, and fail-closed execution rules. |
| 11 | Independent quality reviewer | Fixed Claude CLI `prompt_suggestion` compatibility and source-SQL correlation; no remaining high-confidence quality finding. |
| 12 | Security reviewer | Reviewed path traversal, symlink/TOCTOU, lock replacement, process identity, environment startup controls, redaction, data loss, and CPU/memory bounds. All validated findings fixed. |
| 13 | Mainline / PR drift | Verified the split remains independent of unrelated FK skill content and corrected documentation drift introduced by review fixes. |
| 14 | PR body auditor | Updated scope, review follow-ups, test totals, security gates, and known limitation to match the final diff. |
| 15 | Python idiom reviewer | No additional non-duplicative correctness finding after the fixes. |
| 16 | DSQL authoring-style reviewer | Checked grader choice, assertion wording, corpus placement, and prescribed DSQL terminology. No remaining blocking style issue. |
| 17 | Independent AGENTS.md compliance | Confirmed worktree use, `mise` commands, no new dependencies, clean manifests/cross-references, and full-build compliance. |
| 18 | Reviewer-comment resolution | Classified every review item as fixed, withdrawn, accepted limitation, or non-blocking observation. Eight of eight threads resolved. |
| 19 | Commit / squash audit | Verified intentional commits only, correct merge base, contributor statement, preserved authorship, and no generated/whitespace noise. |
| 20 | Deferred-findings adjudicator | Revalidated disputed findings and promoted only reproducible, customer-relevant issues into the fix set. |
| 21 | Final Code Review convergence | Re-reviewed the complete follow-up diff; no additional high-confidence correctness or repository-instruction finding. |
| 22 | Final security convergence | Found and fixed descriptor-path staging, structural redaction, SQL-comment correlation, deeply escaped credentials, and crash-preparation recovery. |
| 23 | Final tests / comments / types convergence | Confirmed direct regression coverage and documentation consistency after the last security fixes. |

### Validated findings and dispositions

| Area | Finding | Final disposition | Rebased commit |
|---|---|---|---|
| Deterministic grading | Adversarial unbalanced `CREATE TABLE` input caused repeated tail scans | Total scan budget; malformed/budget-exhausted input fails closed | `9477195` |
| Secret classification | Oversized keys and SigV4 `signature` / `sig` / `hmac` handling were incomplete | Linear sensitive-key classification and structured/free-text regressions | `9477195`, `82f175a`, `4476945` |
| Tool correlation | Grader modeled `transact` as `sql` rather than required `sql_list` | Validate nonempty `sql_list`; correlate every statement to source/fixed lint SQL | `82f175a` |
| Lint safety | Unrelated SQL or SQL prefixed with comments could bypass lint gates | Prompt-SQL correlation plus SQL comment canonicalization | `82f175a` |
| Unsafe SQL | Contradictory legacy prose and literal f-string / `%` / `+` construction could pass | Contradiction rejection and structural interpolation detection | `82f175a` |
| Process containment | Reaped/recycled PIDs, lost Linux identity, SIGQUIT, and refresh failures | Root-reaped guard, start-time identity, complete termination-signal handling, fail-closed tracking | `9477195`, `82f175a` |
| CLI compatibility | Unknown-user `expanduser`, leading-dash model values, and post-result prompt suggestions | Resolve errors fail closed, model arguments reject option-like values, prompt suggestions disabled | `9477195`, `82f175a` |
| Environment isolation | Passed startup-control variables could execute before containment | Reject `PYTHONPATH`, loader injection variables, `NODE_OPTIONS`, and equivalent controls | `82f175a` |
| Output concurrency | Replacing the visible lock file allowed separate lock inodes | Lock both output-directory and visible lock-file inodes; continuously verify identity | `82f175a` |
| Staging TOCTOU | Path-based staging could follow a replaced output path | Descriptor-relative staging and artifact writes throughout | `82f175a` |
| Recovery/data loss | Reserved-prefix directories could be deleted without valid runner state or stranded before state publication | Validate journals/markers, preserve unknown entries, use recoverable internal preparation names | `82f175a` |
| Artifact integrity | Environment-secret redaction could rename required schema/control-plane keys and values | Preserve trusted structural keys/enums while redacting untrusted leaves | `82f175a` |
| Redaction completeness | Deeply escaped credentials and truncation boundaries could leak fragments | Forward scanners, deep nested handling, redact-before-truncate ordering | `82f175a`, `4476945` |
| Redaction performance | Escaped JSON, oversized decoded keys, and mapping collisions were superlinear | Linear escaped-key/token scans, collision counters, lazy regex iteration, timing regressions | `82f175a`, `4476945` |
| Documentation | README overstated arbitrary-writer exclusion and sibling cleanup | Document cooperating-runner locks, internal staging only, and exact snapshot scope | `82f175a` |

### Reviewer-specific accounting

- All active blocking requests from `theagenticguy` are implemented.
- The requested total DDL scan budget and SigV4 classification are included.
- Reaped-PID signaling and `expanduser()` failure handling are included.
- The unsafe parent-sibling sweep was replaced with descriptor-relative internal
  staging and recovery.
- Two proposed edits were explicitly withdrawn by the reviewer:
  - hoisting the complete-answer redaction bound into deterministic grading;
  - marker-gating the old sibling sweep.
- The macOS rapid double-fork plus `setsid()` case remains an explicitly
  documented platform limitation, as agreed.
- The non-blocking observation about detecting a context-free 40-character AWS
  secret access key was considered but not implemented because that shape has no
  reliable distinguishing context and would create broad false positives. Named,
  structured, environment, provider-token, access-key-ID, and SigV4 secrets are
  covered.

### Verification

- `mise run test:python` — **116 passed**
- `mise run build` — **passed**
- `git diff --check` — clean
- Build workflow — passed
- CodeQL (`actions`, `javascript-typescript`) — passed
- Dependency Review — passed
- Bandit — passed
- Semgrep / Semgrep OSS — passed
- Gitleaks — passed
- Checkov — passed
- Grype — passed
- ClamAV — passed
- Zizmor — passed
- SonarQube — passed

Only lightweight PR-title, contributor-statement, and merge-status jobs remain
queued; the substantive build and security matrix is green. Human re-review is
still requested from `theagenticguy`.

**Conclusion:** the 20+ agent review fleet is complete, every validated
merge-relevant finding is addressed, and no additional code change is required
from this review.

amaksimo (dismissed):

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,7062 @@
...
+            value,
+            r"f[\"'][^\"'\n]{0,500}"
+            r"\b(?:select|insert|update|delete)\b",
+        )
+        or _has_positive_match(
+            value,
+            r"(?im)^[^\n]{0,1000}(?:sql|query)\s*="
+            r"[^\n]{0,1000}\b(?:select|insert|update|delete)\b"
+            r"[^\n]{0,1000}[\"']\s*(?:\+|%)\s*"
+            r"(?:req\.tenant|[A-Za-z_(])",
```
laithalsaadoon: Blocking, and the same class of bug this revision fixed elsewhere: three nested `[^\n]{0,1000}` gaps with a final `["']\s*(?:\+|%)` that usually never matches, so each line backtracks over the cross product of candidate positions.

Measured on this branch with `"sql="*250 + " select "*125 + '\"'*300 + "\n"` repeated, calling the real function:

| input | time |
|---|---|
| 2.2 KiB | 0.33 s |
| 22.5 KiB | 3.33 s |
| 78.6 KiB | 9.04 s |

Benign prose of the same size is 0.0003 s. That is ~0.115 s per KiB, and the graded text is bounded only by `MAX_CAPTURE_BYTES` (2 MB) at line 4440 — the `MAX_REDACTION_INPUT` guard at 4488 is still judge-only — so one call at the ceiling is several minutes. It runs at 4797 and again at 4812, and grading happens in the parent after the subprocess exits, so neither `--timeout` nor `--judge-timeout` covers it. A run just looks hung.

Gating the gaps with lookaheads keeps the same matches without the backtracking:

```suggestion
            r"(?im)^(?=[^\n]{0,1000}?(?:sql|query)\s*=)"
            r"(?=[^\n]{0,3000}?\b(?:select|insert|update|delete)\b)"
            r"[^\n]{0,3000}?[\"']\s*(?:\+|%)\s*"
            r"(?:req\.tenant|[A-Za-z_(])",
```

After this: **9.04 s -> 0.025 s** at 78.6 KiB, and 0.131 s at 449 KiB. f-string, `+` concatenation, and `%` formatting all still detect True; a parameterised `%s` with `cur.execute(sql, (...))` still returns False. Suite green (51 passed).

Separately, this detector fails open (`passed = not unsafe_interpolation`) and I could bypass it three ways by renaming the variable off `sql`/`query` or splitting the statement across two lines — worth requiring positive evidence of parameterisation rather than absence of a pattern, but that is a design call, not part of this suggestion.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,7062 @@
...
+
+        slash_start = index
+        while index < length and value[index] == "\\":
+            index += 1
+        if index == slash_start or index >= length or value[index] not in "\"'":
+            continue
+        index += 1
+        while index < length and value[index].isspace():
+            index += 1
+        if index >= length or value[index] not in ":=":
```
laithalsaadoon, line 2513: Blocking: this is the one place in the revision where a secret that the previous commit **did** redact now reaches disk. The deleted `ESCAPED_SENSITIVE_VALUE` allowed backslashes before the separator (`\s*\\*[:=]\s*`); the rewrite requires `:` or `=` immediately after whitespace.

Verified by running both the old and new `_redact_text` on the same input:

```
in : prefix {\"password\"\:\"hunter2opensesame\"} suffix
old: prefix {<redacted-secret>} suffix
new: prefix {\"password\"\:\"hunter2opensesame\"} suffix
```

That escaping is what `java.util.Properties` and several log escapers emit, so it is reachable from tool output or a pasted config rather than only hand-crafted text. It lands in `transcript.json`, which is the artifact people actually read and share.

```suggestion
        while index < length and value[index].isspace():
            index += 1
        while index < length and value[index] == "\\":
            index += 1
        if index >= length or value[index] not in ":=":
            continue
```

After this the payload redacts again, the ordinary escaped form still redacts, and the suite stays green (51 passed).

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,7062 @@
...
+
+        slash_start = index
+        while index < length and value[index] == "\\":
+            index += 1
+        if index == slash_start or index >= length or value[index] not in "\"'":
+            continue
+        quote = value[index]
+        index += 1
+
+        closing_end = length
```
laithalsaadoon: Blocking: when no escaped closing quote is found, `closing_end` stays at `length`, `cursor = closing_end`, and the trailing `output.append(value[cursor:])` at 2539 appends nothing — so everything after an unterminated escaped assignment is silently deleted from the redacted string.

Verified against the previous commit, which left this text intact because its lazy `.*?\\+["']` simply did not match:

```
in  (132 chars): assistant said: {\"token\": \"AKIAIOSFODNN7EXAMPLE and then the rest: CREATE TABLE t (id int); lint unfixable
new            : assistant said: {\"<redacted-secret>          <- 96 chars dropped
```

A subject that emits `\"token\": \"` early in its answer erases the rest of its own recorded answer from `transcript.json`, `stderr`, and each timeline `text`. Grading reads raw `text`, so the artifact and the verdict silently disagree.

Leaving the remainder for the later `SENSITIVE_VALUE` / `GENERIC_ASSIGNMENT` passes fixes it:

```suggestion
        closing_end = index
```

After this the access key is still redacted (by the later pass) **and** the trailing answer survives; the ordinary terminated case is unchanged. Suite green (51 passed).

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,7062 @@
...
+                self.lock_descriptor,
+                self.directory_descriptor,
+            )
+            self.lock_descriptor = -1
+        if self.directory_descriptor >= 0:
+            try:
+                fcntl.flock(self.directory_descriptor, fcntl.LOCK_UN)
+            finally:
+                os.close(self.directory_descriptor)
+            self.directory_descriptor = -1
```
laithalsaadoon, line 338: Blocking: the `= -1` assignments come after the calls that can raise, so a failed unlock leaves a stale fd number that a later `close()` (from `__del__` at 341, or `main`'s cleanup loop) closes a second time — by then the number may belong to an unrelated file. `_release_output_lock` already closes the lock fd in its own `finally`, so the same applies to `lock_descriptor` at 333.

Reproduced by forcing `flock(lock_fd, LOCK_UN)` to raise EIO:

```
close() raised: [Errno 5] simulated flock EIO
lock_descriptor field after failure: 4
fd 4 already closed -> Bad file descriptor
new unrelated fd number: 4  collides: True
second close() succeeded (would have closed the recycled fd)
unrelated file BROKEN -> Bad file descriptor
```

Closing the fd already drops the lock, so the unlock can be best-effort and the sentinels should be set first:

```suggestion
        if self.directory_descriptor >= 0:
            descriptor = self.directory_descriptor
            self.directory_descriptor = -1
            with contextlib.suppress(OSError):
                fcntl.flock(descriptor, fcntl.LOCK_UN)
            os.close(descriptor)
```

(Needs `import contextlib`, and the same reordering at 328-333 for `lock_descriptor`.) `DescriptorTemporaryDirectory.cleanup()` at 404 has the mirror-image problem: it clears `entry_name` in a `finally`, so a failed `_remove_output_entry` discards the only record of what to delete, orphaning a mode-0700 `.run-*` directory of transcripts and turning a clean pass into a traceback via the re-raise at 7065.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
```diff
@@ -7,598 +7,7062 @@
...
+
+
+def _redact_artifact_value(
+    value,
+    key: str = "",
+    *,
+    trusted_keys: frozenset[str] = frozenset(),
+):
+    """Apply final sink redaction to every string in persisted output."""
+    if key in TRUSTED_ARTIFACT_VALUE_KEYS:
```
laithalsaadoon: Non-blocking, but worth tightening while you are here: this returns before the `dict`/`list`/`str` branches, so a subtree of any depth under `status`, `artifact_type`, `grader`, `grading_protocol_version`, or `schema_version` is persisted with no value redaction at all.

```
under 'status' : {'status': {'leak': 'password=hunter2', 'nested': ['token=abc123secretvalue']}}
under 'other'  : {'other':  {'leak': '<redacted-aws-credential>', 'nested': ['<redacted-secret>']}}
```

I traced every current producer and there is **no live leak**: all five are harness constants or corpus-validated (`status` is only ever `passed`/`failed`/`error`/`truncated`). But `status` is also in `SAFE_TOOL_RESULT_KEYS` and `_lint_result_is_unfixable` reads a nested `status` out of tool results, so the day someone persists a decoded tool-result dict the subtree ships raw. Restricting the short-circuit to scalars keeps the intent and removes the trap:

```suggestion
    if key in TRUSTED_ARTIFACT_VALUE_KEYS and isinstance(
        value,
        (str, int, float, bool, type(None)),
    ):
        return value
```

laithalsaadoon (dismissed):
Re-reviewed the updated diff at `4476945`. **All five of my blocking findings and the six smaller notes are fixed**, and I verified each one rather than taking the diff's word for it.

What I confirmed resolved, by measurement on this branch:

| finding | evidence |
|---|---|
| quadratic `_create_table_bodies` | 43.51 s -> 0.031 s at 100 KB; 0.125 s at 500 KB; real DDL verdicts unchanged |
| unbounded key into `SENSITIVE_KEY` | 8.14 s -> 0.022 s at 30 KB |
| SigV4 names not sensitive | `X-Amz-Signature`/`sig`/`hmac` now True, `signal`/`design`/`assignment` still False, free text and artifacts both redact |
| sibling sweep in an unowned parent | the parent sweep is gone entirely; staging moved inside the leased dir via `DescriptorTemporaryDirectory`, and a planted `.results.run-user-notes` now survives |
| `killpg` on a reaped PID, stale PID in `signal_known` | `returncode is None` guard added; `signal_known` re-checks `(pid, start_time)` against `/proc` |
| `expanduser` `RuntimeError`, `--model` leading dash, `--pass-env` startup controls, timeout bound | all handled (`_model_argument`, `MAX_TIMEOUT_SECONDS`, `UNSAFE_PASSTHROUGH_ENVIRONMENT`) |

I also checked the new guard tests actually fail when the invariant breaks, rather than passing vacuously. Removing the scan budget, stripping SigV4 coverage from all five places, dropping the `returncode` guard, and reverting the `RuntimeError` handler each produced the matching failure. `test_output_setup_preserves_unowned_sibling_runs` is a nice reversal of the test that blocked my earlier suggestion — that resolved the tension cleanly, and moving staging under the lease is a better answer than the marker-gating I proposed.

That said, the revision is 1,174 changed runner lines, and reviewing the genuinely new code turned up four regressions I would block on, plus one hardening note. Three carry suggestions I applied and measured locally; the suite stays green (`python3 -m pytest -q test_run_functional_evals.py`, 51 passed) with all of them in place.

The one I would look at first is line 3686: the new `_has_unsafe_sql_interpolation` reintroduces exactly the unbounded-quadratic-scan-over-model-text pattern that this revision just fixed in `_create_table_bodies`, and it is reachable from the shipped `safe_query_evals.json`. Since that makes two independent instances, a cheap structural guard may be worth more than fixing them one at a time: bound the deterministic grading input once at the top of `grade_eval` (a `text[:MAX_REDACTION_INPUT]` local for the scanners, keeping raw `text` where anti-truncation matters), or add a test that asserts every registered rule grades a 2 MB adversarial answer within a few seconds.

CI is green on all substantive scans at this head (bandit, semgrep, Semgrep OSS, gitleaks, grype, checkov, clamav, sonarqube, dependency-review, zizmor, build); the four remaining checks are queued housekeeping.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
anwesham-lab: Fixed in f71c3c3 with the suggested lookahead-gated pattern. Added an adversarial timing regression; the 78 KiB reproduction now completes under 1s, and a 460 KiB focused probe completed in 0.05s. Existing f-string/% behavior and the parameterized negative case remain covered.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
anwesham-lab: Fixed in f71c3c3 by consuming escaped backslashes before the separator exactly as suggested. Added a regression for {\"password\"\:\"...\"}; the secret is redacted while surrounding text is preserved.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
anwesham-lab: Fixed in f71c3c3. The unterminated case now leaves the original assignment for downstream SENSITIVE_VALUE/GENERIC_ASSIGNMENT processing instead of consuming the tail. This is the safe equivalent of the suggested closing_end change: access keys and arbitrary single-token passwords remain redacted, while trailing answer/evidence is preserved. Added focused regressions for both cases.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
anwesham-lab: Fixed in f71c3c3. Lock and directory fd sentinels are cleared before fallible operations; unlock is best-effort; each fd is closed once; and DescriptorTemporaryDirectory retains entry_name after failed removal so cleanup can retry. Failure-injection regressions verify recycled fds stay untouched and failed temp cleanup remains recoverable.

tools/evals/databases-on-aws/dsql/scripts/run_functional_evals.py
anwesham-lab: Fixed in f71c3c3 with the scalar-only trusted-value short-circuit exactly as suggested. Added a nested status subtree regression proving password/token values still pass through recursive redaction.

laithalsaadoon (commented):
All five follow-up findings are fixed at `f71c3c3`. Verified by running the real functions on this branch rather than reading the diff:

| finding | before | after |
|---|---|---|
| `_has_unsafe_sql_interpolation` scan cost | 9.04 s at 78.6 KiB | 0.018 s at 78.6 KiB, 0.101 s at 449 KiB |
| escaped separator (`\:`) leak | secret passed through | `<redacted-secret>` |
| unterminated escaped assignment | 96 of 132 chars silently dropped | value redacted, trailing answer preserved |
| `TRUSTED_ARTIFACT_VALUE_KEYS` subtree | nested dict/list written verbatim | nested values redacted, scalar `status` still preserved |
| `OutputDirectoryLease.close()` fd lifecycle | injected unlock EIO left a stale fd that a second `close()` closed out from under an unrelated file | `close()` survives the EIO, both fields land at -1, unrelated fd intact |

Detection is preserved on the regex rewrite: f-string, `+` concatenation, and `%` formatting all still return True, and a parameterised `%s` with `cur.execute(sql, (...))` still returns False.

The twelve new guard tests are not vacuous. I reverted each of the five invariants in place and the matching test failed every time (`test_unsafe_sql_interpolation_scan_is_bounded`, `test_escaped_sensitive_assignment_allows_escaped_separator`, `test_unterminated_escaped_assignment_preserves_trailing_answer`, `test_unterminated_escaped_password_redacts_value_and_preserves_tail`, `test_trusted_artifact_value_bypass_applies_only_to_scalars`). Suite is 60 passed on a clean tree.

## Correction to my severity ranking

My previous review led with the scan-cost finding and called it the thing to fix first. That ranking was wrong, and I want it on the record before this merges so the next reader does not inherit it.

I had assumed a CI or service threat model. This harness is neither. No workflow invokes `run_functional_evals.py`; CI only runs its unit tests via `pytest tools plugins`. Every documented invocation writes to `/tmp`, no raw artifact is committed anywhere in the repo, the corpus is in-repo under a CODEOWNERS rule covering `*`, and reaching a live cluster additionally requires the operator's own credentials plus an `--mcp-config` the README already labels as needing review. There is no multi-tenancy, no shared runner, and a human in front of every run.

Under that model:

1. The transact guard is still the sharpest surface. Real credentials against a real cluster, write tool allowlisted, denial delegated to a hook. The `shlex.quote` fix in the prior revision addressed my specific complaint, and this context confirms the concern rather than softening it.
2. The transcript truncation bug should have been ranked above the scan cost. The transcript is the product: a maintainer runs an eval, it fails, they read `transcript.json`. Silently deleting the tail while grading used the full text means the artifact and the verdict disagree with no marker, in the one file a human will read.
3. The two redaction gaps hold at medium, but for a narrower reason than I gave. There is no live exfiltration path. What they defeat is the README's own control, which is a human reviewing artifacts before publication, and a plain `X-Amz-Signature=<64 hex>` is what a skimming reader misses.
4. The fd lifecycle bug is local and single-process. Low.
5. The scan cost is hygiene, not a blocker. Triggering it needs the subject model to emit several hundred kilobytes of adversarially shaped text in response to a reviewed prompt, and the cost is a maintainer waiting and pressing Ctrl-C. Worth bounding, which you did; not worth blocking on.

Bounding the deterministic grading input once in `grade_eval` is still worth doing as hygiene, since this was the second instance of the same unbounded-scan shape. `test_unsafe_sql_interpolation_scan_is_bounded` and `test_create_table_body_scan_is_bounded` do run in CI, so the class now has regression coverage either way.

Not approving yet only because CI for `f71c3c3` is still queued. No outstanding code objections from me.

anwesham-lab (approved):

laithalsaadoon (approved):
LGTM

PR #263, opened 2026-08-28
docs(dsql): require official Flyway module

amaksimo:
Add Aurora DSQL Flyway module-selection and migration guidance to the DSQL skill.

#### Related

- https://github.com/flyway/flyway-community-db-support/pull/139
- https://github.com/awslabs/aurora-dsql-tools/pull/150

#### Changes

- route Flyway setup and migration questions to a focused reference
- always select the official `org.flywaydb:flyway-database-dsql` community module
- never select or recommend the deprecated `software.amazon.dsql:aurora-dsql-flyway-support` adapter
- direct existing adapter users through a tested dependency swap that preserves migration state
- prohibit loading both modules in one Flyway runtime
- add a functional eval covering official-module selection, migration, coexistence, and schema-history preservation

Verification: `mise run build`; focused eval 17 passed 6/6 expectations.

#### Acknowledgment

### DIFF
--- a/tools/evals/databases-on-aws/dsql/evals.json
+++ b/tools/evals/databases-on-aws/dsql/evals.json
@@ -201,14 +201,14 @@
     },
     {
       "id": 17,
-      "prompt": "We already run Aurora DSQL migrations with `software.amazon.dsql:aurora-dsql-flyway-support`. The official `flyway-database-dsql` PR is merged, but I cannot find that artifact in Maven Central yet. Can I add the new module alongside the AWS adapter to try it? Give me the safest migration path without breaking our existing schema history.",
-      "expected_output": "Treats the deprecated AWS adapter as a maintenance-only continuity path for existing users, waits for a published upstream artifact compatible with the user's Flyway runtime, never loads both adapters together, and gives a non-production dependency-swap validation sequence that preserves existing migration state.",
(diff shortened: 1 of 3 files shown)


tools/evals/databases-on-aws/dsql/evals.json
```diff
@@ -198,6 +198,21 @@
         "Routes the query to Workflow 9 (EXPLAIN ANALYZE / query plan explainability) to establish the actual plan and root cause, and does not recommend index or schema changes from AAS data alone"
       ],
       "llm_judge": true
+    },
+    {
+      "id": 17,
+      "prompt": "We already run Aurora DSQL migrations with `software.amazon.dsql:aurora-dsql-flyway-support`. Should we keep using it or add the official `flyway-database-dsql` module alongside it? Give me the safest migration path without breaking our existing schema history.",
```
anwesham-lab, line 204: seems like a prompt eval that might be too guiding if it's the only one? Should we add a check against an explicit ask of how can I use Flyway to migrate to DSQL? Or something similar?

Can we check how it would have responded without the flyway guidance at all? What was the baseline against these prompts?

anwesham-lab (commented):
Unclear to me if we want this full descript rather than explicitly pointing it to a DSQL Flyway readme of sorts?

PR #275, opened 2026-09-17
feat(elastic-beanstalk): add Beanstalk Cluster mode

karancode:
Adds Beanstalk Cluster mode to the existing `elastic-beanstalk` skill. Cluster mode runs a containerised application on a managed Amazon EKS Auto Mode cluster. The skill previously covered Standard mode on EC2 only and did not say so, so an agent asked for a Cluster deployment produced Standard mode configuration.

This extends the existing skill rather than adding a new one, so mode selection sits with the rest of the Elastic Beanstalk routing.

#### Related

- [Configuration options for Beanstalk Cluster environments](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/command-options-general-eks.html)
- [Scaling Beanstalk Cluster environments](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/configuring-cluster-scaling.html)

#### Changes

New file:

- `skills/elastic-beanstalk/references/cluster-mode.md` — the twelve `eks` namespaces, required IAM roles, environment creation, defaults worth setting deliberately, subnets, HTTPS, multi-service discovery and ingress groups, autoscaling, verification, troubleshooting order, immutable settings, cost

Modified:

- `skills/elastic-beanstalk/SKILL.md` — mode routing table, Cluster mode description triggers, existing content scoped as Standard, reference link
- `skills/elastic-beanstalk/references/platforms.md` — scope note: Cluster mode has no platform
- `skills/elastic-beanstalk/references/configuration.md` — scope note: none of those namespaces apply to Cluster mode

The mode boundary is stated firmly because a Cluster environment accepts twelve `aws:elasticbeanstalk:eks*` namespaces and rejects a classic namespace with `InvalidParameterValueException` rather than ignoring it, so Standard mode configuration fails at create.

Two notes for reviewers. `SKILL.md` goes from 172 lines ("ideal") to 215 ("good"); holding "ideal" would mean moving existing Standard content into `configuration.md`, which I left alone — happy to do it if you would prefer. And the parent `deploy` skill still routes generic containerised workloads to Fargate, so an implicit request that never names Elastic Beanstalk may not reach this mode table; that is a service-selection change and belongs in its own PR.

Validation: `dprint check`, `markdownlint-cli2`, `validate-references.py`, `validate-cross-refs.cjs`, `validate-size.py` and `validate-urls.py` all clean. Also exercised the skill via `--plugin-dir` against three prompts — explicit Cluster request, ambiguous single container, and multi-service — to confirm mode routing and that no Standard-mode namespaces appear in Cluster output.

#### Acknowledgment


plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
```diff
@@ -0,0 +1,418 @@
...
+```
+
+Three API calls, and the tier is what selects Cluster mode:
+
+```bash
+aws elasticbeanstalk create-application --application-name my-app
+
+aws elasticbeanstalk create-application-version \
+  --application-name my-app --version-label v1 \
+  --image-source Uri=public.ecr.aws/my-org/my-service:1.0.0
```
amjadsy: `--image-source` is not a valid `CreateApplicationVersion` parameter in the launched CLI model, so this command fails before an application version is created. The image must be nested under `ImageConfiguration.Source`, for example:

```bash
--image-configuration "Source={Uri=public.ecr.aws/my-org/my-service:1.0.0}"
```

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
```diff
@@ -0,0 +1,418 @@
...
+  --application-name my-app --environment-name my-app-env \
+  --query "ConfigurationSettings[0].OptionSettings[?OptionName=='cluster-role' || OptionName=='node-role' || OptionName=='observability-role'].[OptionName,Value]" \
+  --output table
+```
+
+An image-based application version reports `Status: UNPROCESSED` and is
+immediately deployable. Only source bundles are processed, because only they
+need building. To deploy source instead of an image, supply a source bundle and
+Elastic Beanstalk builds the image with AWS CodeBuild, from a Dockerfile if
+there is one or from a buildpack if there is not.
```
amjadsy: This currently implies that Beanstalk detects Dockerfile versus buildpack automatically, but source-based Cluster versions require an explicit source-build request: `--source-bundle`, `--process`, and `ImageConfiguration.Build`. `Build.Type` must be `docker` or `buildpack`, `CodeBuildServiceRole` is required, and a buildpack build must name its builder. Without this, the version remains `UNPROCESSED` or creation fails. Please provide a complete example or link to the source-build procedure and warn against reusing the Standard-mode source workflow.

plugins/deploy-on-aws/skills/elastic-beanstalk/SKILL.md
```diff
@@ -56,7 +60,33 @@ Lambda only when the user explicitly asks for serverless or the workload is
...
+Decide before generating anything. The modes share an API and nothing else: a
+Standard mode namespace passed to a Cluster mode environment is rejected, not
+ignored.
+
+| Signal                                                      | Mode         |
+| ----------------------------------------------------------- | ------------ |
+| Source on a standard runtime; Windows or .NET Framework     | **Standard** |
+| Worker consuming an SQS queue                               | **Standard** |
+| One small application, cost is the priority                 | **Standard** |
+| Several services needing private service-to-service traffic | **Cluster**  |
```
amjadsy, line 87: Please gate this recommendation on the Cluster application contract. Cluster requires one or more identical, interchangeable replicas, and local storage is ephemeral. A multi-service signal alone can therefore route an incompatible stateful application here. Add a rule such as: use Cluster only when every service can run as interchangeable replicas and all durable state, uploads, and sessions live outside the container. The earlier “local state” statement should also be scoped to Standard mode.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
```diff
@@ -0,0 +1,418 @@
...
+The value is a comma-separated list, which the CLI shorthand form reads as a
+field separator, so pass subnets in a JSON file rather than on the command line.
+
+## HTTPS Only
+
+Elastic Beanstalk configures an HTTPS listener on port 443 and leaves port 80
+closed. Always verify over `https://`. An `http://` request hangs until it times
+out and reads as a broken deployment.
+
+Elastic Beanstalk creates and attaches a certificate for the environment's own
```
amjadsy: This is only true for a Beanstalk-managed ALB. The guide also exposes the `aws:elasticbeanstalk:eks:alb` `arn` option; when the user supplies an existing ALB, Beanstalk does not create its HTTPS listener or certificate, and the customer owns the listeners, TLS configuration, and scheme. Please branch this section so an agent does not assume HTTPS is already configured for a customer-owned ALB.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
```diff
@@ -0,0 +1,418 @@
...
+2. **Health** (`describe-environment-health --attribute-names All`) — the
+   `Causes` field. `ELB health is failing or not available for all instances.`
+   means the load balancer cannot get a healthy response.
+3. **Target group** (`elbv2 describe-target-health`) — the port and the reason.
+   `Target.Timeout` on a port is the strongest single clue available.
+4. **Replica count** — a steady count means a pod is running and staying
+   running, so the application is alive and unreachable rather than crashing.
+5. **Application logs** — the only place the application itself speaks.
+
+Logs go to three account-wide log groups, shared across every environment rather
```
amjadsy: The monitoring contract defines four account-wide CloudWatch log groups, not three: application logs, application metrics, infrastructure logs, and infrastructure metrics. Application log streams use `eb-<environment-name>.<pod-name>`, not the longer deployment pattern below. The tail command remains usable because its prefix is broad, but the explanation should match the actual contract. It is also worth noting that these groups are created without a retention policy.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
```diff
@@ -0,0 +1,418 @@
...
+
+Elastic Beanstalk adds no service fee. A Cluster mode environment costs:
+
+- The Amazon EKS cluster's hourly charge.
+- The EC2 instances that EKS Auto Mode provisions to hold the replicas, **plus
+  the EKS Auto Mode management charge on those instances**, which is billed per
+  instance on top of the instance price and varies by instance type.
+- An Application Load Balancer per environment, unless `load-balancer-type` is
+  `None`.
+- EBS volumes for any persistent storage, plus data transfer.
```
amjadsy: Cluster currently exposes no persistent application-volume or PVC configuration. EKS Auto Mode block storage being enabled for the cluster does not give an environment managed persistent storage, and local container storage is explicitly ephemeral. Please remove this bullet and say that durable state must live in an external service managed separately; do not imply that EBS or an EFS mount is available through the Cluster configuration surface.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
```diff
@@ -0,0 +1,418 @@
...
+| `max-replica`                                     | `10`                              | Ten replicas is a cost surprise, not a failure. Pin it to what the workload needs.             |
+| `aws:elasticbeanstalk:eks:alb` `healthcheck-path` | `/`                               | `/` is rarely a real health endpoint and often does real work. Point it at one that is.        |
+| `subnets`                                         | public subnets of the default VPC | Always set these. See [Subnets](#subnets) — the default is public and cannot be changed later. |
+| `readiness-probe` `enabled`                       | `false`                           | All three container probes are off until enabled.                                              |
+
+Two more worth knowing rather than setting. `service-port` defaults to `8080`
+and is injected into the container as `PORT`, overriding any `PORT` the user
+supplies in `env-variables`. And `env-variables` is a single option holding a
+JSON object of every variable, not one option per variable — the same is true of
+`secrets`.
```
amjadsy, line 270: Please add a dedicated secrets section rather than leaving `secrets` as a JSON-shape note. Using `secrets` requires `application-role`, a Pod Identity trust policy for `pods.eks.amazonaws.com`, and permission to read every referenced Secrets Manager or Parameter Store value; without the role, the volume mount fails and replicas do not start. The same prerequisite should be connected to `scaler-auth-secret` and observability `custom-credentials`, since the parent skill tells agents not to inject secret values directly.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
```diff
@@ -0,0 +1,418 @@
...
+because it needs an HTTP listener that the environment does not have by default;
+add one through `listen-ports` if the user wants the redirect.
+
+## Multiple Services
+
+A Cluster mode environment runs one service. An application made of several
+services is several environments, one per service, and each can carry its own
+Elastic Beanstalk application so it keeps its own version history.
+
+Environments that share subnets land on the same EKS cluster, which is what lets
```
amjadsy, line 380: Please state the security boundary before recommending subnet-based cluster sharing. Environments on the same cluster use soft multi-tenancy and share the EKS control plane and, by default, worker nodes. Sharing is appropriate for services one team owns and operates together; different end customers, untrusted code, compliance boundaries, and commonly production versus development should use different subnet sets and therefore separate clusters. A unique `node-pool` provides dedicated nodes, but it does not create a separate control plane or failure domain.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
```diff
@@ -0,0 +1,418 @@
...
+schedules the shared cluster for deletion a few hours after the last environment
+on it terminates.
+
+Cluster mode maintains nodes and keeps its own cluster add-ons current. It does
+not patch the application runtime or the container image, so keeping the image
+current stays with the user.
+
+## Cost
+
+Elastic Beanstalk adds no service fee. A Cluster mode environment costs:
```
amjadsy, line 604: The estimate needs to include the material default costs outside this list: Cluster metrics use three paid custom CloudWatch namespaces and scale with replica count; four shared log groups are created without retention; private-subnet deployments may require a NAT gateway or VPC endpoints; and source builds incur CodeBuild and ECR costs. Please carry these into the parent cost-estimation guidance so mode selection does not compare Standard against only the EKS, node, and ALB line items.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
karancode: Fixed in bc6c7d6 — now `--image-configuration Source={Uri=...}`. You are right, and the command as written failed before creating anything.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
karancode: Fixed in bc6c7d6. There is now an Application Versions section, because the auto-detection claim was wrong in a way that needed more than a corrected sentence. It states that a source build is never automatic and nothing is inferred from the bundle, gives the four required pieces as a table (`--source-bundle`, `--process`, `Build.Type`, `Build.CodeBuildServiceRole`), notes that `--process` omitted leaves the version `UNPROCESSED` with no build started, requires the `Buildpack` builder for a buildpack build, covers the `Architecture`/`arch` match, and covers polling `BUILDING` to a terminal status with `BuildArn` for the CodeBuild logs. Also an explicit "do not reuse the Standard mode source workflow" paragraph naming `create-storage-location`, solution stacks and `--build-configuration`, plus the mutual-exclusion rules for `ImageConfiguration`. Linked to beanstalk-cluster-app-versions.html.

plugins/deploy-on-aws/skills/elastic-beanstalk/SKILL.md
karancode: Fixed in bc6c7d6. Added an application contract to cluster-mode.md — interchangeable replicas, ephemeral local storage, state outside the container — and gated mode selection on it in SKILL.md: it is checked before the signals table, and explicitly does not yield to a multi-service signal. Added a "writes uploads, sessions, or working files to local disk" → Standard row, and scoped the Lambda contrast to "in Standard mode only" for local state.

Verified with a smoke test: a Django + Celery app with local-disk uploads and file sessions, asking for private service-to-service traffic, now routes to Standard and names the contract as the reason. It went to Cluster before this change.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
karancode: Fixed in bc6c7d6. The section now opens with a three-row table — Beanstalk-managed ALB, `arn` set to an existing ALB, `load-balancer-type: None` — and says up front that the rest applies only to the first. For a supplied load balancer it states that the listeners, TLS and scheme are the user's, that `certificate-arn`, `listen-ports`, `ssl-redirect`, `scheme` and the `:alb` `subnets` should not be set expecting them to take effect, that the agent should check what the load balancer actually has before telling the user a URL to test, and that an NLB ARN is rejected. Also added the `listen-ports` JSON shape while I was there.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
karancode: Fixed in bc6c7d6 — four groups, as a table with the correct stream name per group. Called out the difference you are pointing at, since it is the part that misleads: application logs are `eb-<environment-name>.<pod-name>` while application metrics are `<environment-name>/<pod-name>` with no `eb-` prefix. Added the no-retention-policy note, and that a new set of streams appears on every deployment and scaling event, with a suggestion to set retention on all four — which also feeds the cost section.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
karancode: Fixed in bc6c7d6 — bullet removed. This was an inference on my part from EKS Auto Mode block storage, and my own namespace inventory contradicts it, which I should have noticed. cluster-mode.md now states that there is no persistent volume or PVC surface, that EKS Auto Mode block storage being available to the cluster does not give an environment a managed volume, that durable state is an external service managed separately, and not to imply an EBS or EFS mount is reachable through Cluster configuration. It is in the application contract section, so it also feeds the routing gate.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
karancode: Fixed in bc6c7d6 — there is now a Secrets and the Application Role section. It opens by restating the parent skill's rule, then covers the three things that have to be set up together: the role created before the environment, the Pod Identity trust policy allowing `sts:AssumeRole` and `sts:TagSession` for `pods.eks.amazonaws.com`, and read permission on every referenced value. It states the failure mode plainly — the mount fails and the replicas never start, so the environment does not come up at all — and keeps the `GetSecretValue`/`DescribeSecret` point, since with only the first it starts and then fails on a later refresh.

A table connects the three options that share the prerequisite: `secrets`, `scaler-auth-secret`, and observability `custom-credentials`. The autoscaling section now points here instead of repeating it, and the required-roles section flags the application role as required in practice by more than its description suggests.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
karancode: Fixed in bc6c7d6. The security boundary now comes before the sharing recommendation. Names it as soft multi-tenancy with a shared control plane and, by default, shared nodes, and says logical separation is not infrastructure separation. A requirement-to-boundary table sends different end customers, untrusted code, compliance boundaries, and production-versus-development to different subnet sets, and says the cost difference is the trade being made rather than resolving it silently in favour of the cheaper option. Also added the three things a shared cluster does not separate: egress is not restricted at all and no option restricts it, the control plane and its Kubernetes version are shared, and cluster drift fails updates for every environment on it.

On `node-pool`: kept as the answer to a node-capacity or noisy-neighbour requirement, explicitly not a security boundary, with the detail that environments sharing a `node-pool` value share nodes with each other — so a value another environment already uses is not dedicated.

plugins/deploy-on-aws/skills/elastic-beanstalk/references/cluster-mode.md
karancode: Fixed in bc6c7d6, in both places. cluster-mode.md now lists all four as defaults that are easy to miss because none of them appear in the environment's configuration: the three paid custom namespaces scaling per replica, the unretained log groups, NAT or VPC endpoints for private subnets, and CodeBuild plus ECR for source builds.

Carried into `skills/deploy/references/cost-estimation.md` as you asked, with a Beanstalk Cluster subsection: a line-item table, the service codes for EKS, CloudWatch, NAT and CodeBuild, and the point that the custom metrics are a *new* line item rather than a larger one, since Standard publishes to `AWS/ElasticBeanstalk` free. It also states that only the cluster hour is shared and that one small service costs more in Cluster mode than on a single EC2 instance, so the comparison is not made on EKS, nodes and ALB alone.

Flagging that this is the one change outside the EB skill — say the word and I will split it into its own PR.

karancode:
All nine comments addressed in bc6c7d6, thanks @amjadsy.

krokoko:
Hi @karancode , could you please bump the plugin version in all required places ? Thanks ! (You can check past PRs)

karancode:
@krokoko thanks! Done in 69fff1c — `deploy-on-aws` 1.3.0 → 1.4.0 (minor, since Cluster mode is a new capability), in all three places: the marketplace entry, `.claude-plugin/plugin.json`, and `.codex-plugin/plugin.json`.

I hand-edited the Codex manifest rather than regenerating it, because `tools/generate_codex_manifests.py` fails on `main` before it reaches `deploy-on-aws`:

```
ValueError: Missing interface metadata for plugin 'codebase-documentor-for-aws'.
```

It also rewrites an unrelated keyword in `plugins/aws-serverless/.codex-plugin/plugin.json` (drops `capacity-provider`) on the way, which I reverted rather than carry here. Both look like pre-existing issues worth a separate fix — happy to open an issue if useful.

krokoko (approved):

